PR #1, opened 2016-07-28
matrix only impl

michaelsbradleyjr:
I noticed that generation of categorical data was rather slow compared to the generation of other distributions.  This pull request represents my attempt to make categorical generation much faster, by using only matrix operations, i.e. without `data.table` subsetting (cf. [line #30](https://github.com/kgoldfeld/simstudy/blob/master/R/int_gencat.R#L30) in `master`).

With the changes in place, generation time is dramatically reduced, e.g. for sample sizes in the millions.

Let `sim.dt` be a table returned from `simstudy::genData`; I spot-tested whether the categorical distribution conformed to the specified probabilities like so:

```R
fact <- as.factor(sim.dt$xCat)
smry <- summary(fact)
smry/sum(smry)
```

And indeed that appeared to be the case. Let me know if/how I can make this pull request better, and I will be happy to do so.

### DIFF
--- a/R/int_gencat.R
+++ b/R/int_gencat.R
@@ -9,39 +9,20 @@
 
 gencat <- function(n, formula, dfSim) {
 
-  # 'declare var
+    # parse formula
 
-  V1 = NULL
+    pstr <- unlist(strsplit(as.character(formula), split = ";", fixed = TRUE))
 
-  #
+    # build command and parameters
 
-  pstr <- unlist(strsplit(as.character(formula),split=";", fixed=TRUE))
-  idname <- names(dfSim)[1]
+    cmd <- substitute(
+      (t(stats::rmultinom(nsamp, 1, probs)) %*% c(1:nparam))[,1],
+      list(nsamp  = n,
+           probs  = as.numeric(pstr),
+           nparam = length(pstr))
+    )
 
-  dtSim <- data.table::data.table(dfSim)
+    # evaluate and return
 
-  ps <- paste0(pstr, collapse=",")
-  ps <- paste0("c(", ps, ")")
-
-  nparam = length(pstr)
-
-  # build command based on parameters "ps"
-
-  cmd  <- quote(x[, 1]) # x is [[2]]
-  mcmd <- quote(x %*% c(1:nparam)) # x is [[2]]
-  tcmd <- quote(t(x)) # x is [[2]]
-  pcmd <- quote(stats::rmultinom(nrow(dtSim), 1, x)) # x is [[4]]
-
-  pcmd[[4]] <- parse(text=ps)[[1]]
-  tcmd[[2]] <- pcmd
-  mcmd[[2]] <- tcmd
-  cmd[[2]]  <- mcmd
-
-  # if (!all(apply(p,1,sum) == 1)) {
-  #   stop("Sums for cumulative probabilities in categorical distribution not 1")
-  # }
-
-  new <- eval(cmd)
-
-  return(new)
+    return(eval(cmd))
 }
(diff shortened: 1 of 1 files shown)


kgoldfeld:
I have been aware of this problem for a while and meant to fix it. I appreciate your suggested fix. I will check it out and make the changes soon. Thanks.

Generally, was wondering how find the package.

michaelsbradleyjr:
Okay, great to hear.

I just started working with the package yesterday, and studying how it works.  I think it's going to work a lot better than the simulation-generator code I cooked up in recent months. Of course, when I started working on my project, I was new to R and my early work on simulations very much reflects that.

I don't know if you'd be interested, but I cooked up a function and a string that I'm copy-pasting for working with tables that I'm feeding to defRead:

```R
# vartable.string <- {'
#   varname ,       formula , variance ,        dist ,     link
#   --------|---------------|----------|-------------|---------';'
#        nr ,             7 ,        0 ,   nonrandom , identity
#        x1 ,         10;20 ,        0 ,     uniform , identity
#        y1 ,    nr + x1 *2 ,        8 ,      normal , identity
#        y2 , nr - 0.2 * x1 ,        0 ,     poisson ,      log
#      xCat ,   0.3;0.2;0.5 ,        0 , categorical , identity
#        g1 ,      5 + xCat ,        1 ,       gamma ,      log
#        a1 ,     -3 + xCat ,        0 ,      binary ,    logit
# '}

defRead.string <- function(str) {
  fields     <- 'varname, formula, variance, dist, link'
  csv.file_1 <- tempfile(pattern = "vartab", tmpdir = tempdir(),
                         fileext = ".csv")
  write(paste(fields, str), file = csv.file_1)
  csv.file_2 <- tempfile(pattern = "vartab", tmpdir = tempdir(),
                         fileext = ".csv")
  write.table(data.table::fread(csv.file_1), file = csv.file_2, sep = ",",
              quote = F, row.names = F)
  simstudy::defRead(csv.file_2)
}
```

I'll likely implement a `defReadAdd.string` function as well.  When I want to create a new generator table, I just copy-paste the commented code, uncomment, and edit. I find it to be a happy medium between using `defData` vs. an CSV file.

kgoldfeld:
Michael -

I haven't totally disappeared - just got sidetracked with other things. I do want to incorporate these changes, or at least address the problem. I do have to admit that I am a newbie to GitHub, and I don't want to totally mess things up with the Pull request, so I am going slowly.

- Keith

kgoldfeld:
Michael -

I've finally taken a look at your solution. Definitely faster, but it has a drawback of not being quite flexible enough to handle the situation where probabilities are a function of pre-defined variables. The approach you suggested works only if the probabilities are defined as constants.

I will continue to look into this.

- Keith

michaelsbradleyjr:
Thank you.

kgoldfeld:
I've found rMultinom in package Hmisc that allows the specification of the probabilities as a matrix - this would speed up the function by a factor of 4 - not nearly as good as it can be with rmultinom from package stats. Do you think this would be acceptable? Perhaps I can check to see if the probabilities are constant across records, in which case rmultinom would be used - otherwise rMultinom would be used. What do you think?

michaelsbradleyjr:
Yes, I did not properly consider probabilities being given as expressions involving previously defined variables. I was fairly green with `simstudy` when I made this pull request, and have not had cause to reconsider it. I think doing a check as you propose, and selecting the faster method if they're all constant, sounds like a good solution.

kgoldfeld:
Just wanted to provide one more update. I did indeed speed up gencat, but used an Rcpp function I wrote earlier to observed treatment assignment, which is essentially the same thing as multinomial number generation. The final solution is a tad slower than what you proposed (but not much) and provides flexibility for setting the probability levels.

If you continue to use simstudy and have further suggestions, questions or anything else, let me know. I am curious to hear how useful it is. (I will take a look at your other suggestion at some point.) I know it is far from perfect.

PR #8, opened 2019-04-15
Fix .genQuantU command to include corMatrix.

abeeisnotabug:
Changed the corMatrix argument in the .genQuantU command to accept the corMatrix provided by the user.


kgoldfeld:
Another user just contacted me about this issue - very strange co-incidence. I will make the necessary change.

assignUser:
Thank you for your pr, the fix was already implemented :)

PR #9, opened 2019-08-30
Bugfix incorrect lower bound calculation in .checkBoundsBin() function

gsverhoeven:
Hi!
While playing around with your nice package, I ran into a bug with `.checkBoundsBin()`
The lower boundary was not calculated correctly because of missing parentheses in the denominator.

A colleague of mine suggested to add a unit test for this function as an example of how easy it is to create a few unit tests when developing code. This is my first unit test ever, but i found he was right, it is really straightforward. It uses the ``testthat`` framework that is widely used for R packages.

Using devtools::install(), the package compiles, runs the tests, and installs without problems on my machine.

regards,
Gertjan

### DIFF
--- a/R/int_binEPmethod.R
+++ b/R/int_binEPmethod.R
@@ -18,7 +18,7 @@
       
       if (d < L | d > U)   {
         LU <- paste0("(", round(L,2) , " ... ", round(U, 2), ")")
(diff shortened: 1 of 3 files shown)


kgoldfeld:
Thanks for the catch, and for reminding me that I really do need to add unit tests at some point.

PR #12, opened 2020-04-11
Pseudo-random distribution & documentation

assignUser:
Hello!

Thank you for your package! I had a use case not covered by your package which I had to manually add to the data. It looked so ugly that I went ahead and implemented a solution! 😄  

I added two new distributions: `pseudorandom` and `pseudorandomSeq` these are useful if you want to assign a number of values in random order, a set amount of times (e.g. sensory stimulations). In contrast to `uniform` and `uniformInt` which give _about_ the same number of times per value but not exactly.

I added the new distributions and integrated them into .generate() with some tests and documentation.

Additionally I added a help page that gives an overview about all possible distributions and their expected values. One can find these in the vignette but its spread out over a page and not very structured (sorry ;) ), some where missing (mixture, exponential) completely afaik.


assignUser:
Is simstudy.R ceated automatically during the build? Or did I accidentaly change it? If so I will of course revert the changes.

assignUser:
Seem to have found an issue, going to write some tests.

assignUser:
Found the issue, was working with the internal functions directly, which is of course bad form but also skipped .defEval. I have created some test that check user Workflow and adjusted defEval.

kgoldfeld:
Thanks for doing this - I'll have time to take a look tomorrow.

vignettes/simstudy.R is created automatically during build, but I was having an issue with the vignette a couple of versions ago - CRAN would not accept it. So, I basically gutted in down to the essentials and put more detail on my website (rdatagen.net).

assignUser:
Sounds great!

If you don't mind me intruding I have some more ideas for the package :D I will create some Issues so we have a place to discuss them. 

For now happy Easter/ happy Pesach!

assignUser:
This will need adjustments for the new formula style and we need to find a good name.

PR #22, opened 2020-05-06
New formula system & more tests

assignUser:
As we talked about this is not release ready (Documentation/Vignette need updates, examples, more tests...) but everything should work and be backwards compatible.
Closes #13, closes #14, fixes #15, fixes #17

Issue #18, opened 2020-04-19
defData: first row not checked


assignUser:
While getting the new evalDef working (need to pass link from defData), I noticed that the first defined variable is not checked by evalDef, only tested to see if it is scalar.

kgoldfeld:
The first call to a defData (when creating a new definition) is creating the first variable, there is no way to reference an existing variable in the data.table. So, the formula can only be a scalar. (This is not true for defDataAdd, where by definition that is used to add data to an existing data.table.) But, now that I think about it, I guess all the checks that are done elsewhere could be done in the first variable, except we need to check that no variable was referenced as well. Does that make sense?


I never liked the fact that I had defData and defDataAdd. But I did it because in defData I could check the definition as it was being built. In the case of defDataAdd, that was not possible, so I only checked that when the definition was being executed. I guess I should have done defData the same way, and could have had a single function.

________________________________

assignUser:
Yeah I understand why you did it this way, but we still can check formula formats etc. :) 

I will hot fix defData for now to get the new evalDef working and testing. Afterwards I'll look at it and defDataAdd. Maybe we can unify them a bit,

assignUser:
One idea I have to unify defData/Add would be to add a `dt` argument to defData which is used to check for previously defined vars in case it is given. This would of course change the api in a backwards incompatibily way...
Alternativly we could add a flag `add = TRUE/FALSE` that disables the previously defined checks.

Let me know if you have any other ideas :)

assignUser:
Or the `add` flag could be either a boolean or a dt, in case it is dt we can check validity but in case of bool we just don't check (as is the behavior of `defDataAdd`no).

This would allow for backward compatibility while also allowing for more "security" with adding new vars.

kgoldfeld:
This is in line with what I had been thinking. In concept, I like the 2nd option you just mentioned. But, as I think about it, one issue I see is users (me, for example), might like to define all the data generating processes before actually generating the data - so that it would be impossible to specify the data set:

```{r}

# Define data

d1 <- defData(...)
...
d1 <- defData(d1, ...)

d2 <- defDataAdd(...)
...
d2 <- defDataAdd(d2, ...)

# Generate data

dd <- genData(1000, d1)
dd <- addColumns(d2, dd)
```

assignUser:
Maybe im just a bit dumb right now but why would you need two definitions in that case?

Addtionaly the "dual" flag would allow for just setting it to TRUE in that case. 
[...]

Issue #15, opened 2020-04-13
Uniform/Int distribution does not check formula


assignUser:
There is no error, just produces NAs.
```r
 def <- defData( varname = "x1", dist = "uniformInt", formula = "30;20")
genData(5,def)
 
   id x1
1:  1 NA
2:  2 NA
3:  3 NA
```

kgoldfeld:
Thanks - that needs to be fixed - I'll make sure to flag this an error, since min cannot exceed max.  I'll wait until you find more errors.

I'd love to hear how you are developing these tests ...

assignUser:
Sorry for the late reply, I'm pretty into it right now but gave into my usual weakness. Namely starting big not small, so I have nothing finished yet xD I will show you soon though.

You can mark this issue with the "bug"/"error" label. If I find any additional issues I will create new ones, with the intent of keeping a record and fixing them later, not pawning them of to you!

kgoldfeld:
Sounds good. I don't mind fixing the bugs, but am very happy to have you work on it.


Are you getting credit for this, or just interested in getting involved? As you can probably tell, I am not a professional programmer, and there are decisions I made early on that I wish I could change, but unfortunately cannot given that there are thousands of users out there. In any case, I am very happy to have someone to help me fix, improvi, and grow this package.

________________________________

assignUser:
I just noticed that I ruled out variables in formulas for the uniform/Int and categorical distributions with the way I wrote the new evalDef. Previously as far as I can see this should be possible, so I will change it to keep the same functionality!

As an update: I am currently working on test coverage for the generator functions and modifying them to work with external variables.

assignUser:
For consistency I have added the same behavior to .gencat for probabilities != 1 that catProbs uses.

assignUser:
Closed by #22

PR #24, opened 2020-07-19
Ensure correct dist values.

vlulla:
Validate dist name at data definition instead of waiting until data generation.


assignUser:
Thank you for your contribution! We are currently working on a complete overhaul of the formula and definition system ( #22 #18 #15) that also addresses the issue you raised here. Consequently we are not going to merge your pr but we appreciate your input! 😃

assignUser:
@vlulla as of 0.2.1  this should now happen as you requested

vlulla:
Thank you!

PR #29, opened 2020-08-10
Add actions & pkgdown

assignUser:
I have added the workflow files we discussed with additional files that allow us to manually trigger R-CMD-CHECK and the generation of the pkgdown page (we probably want that to be on par with the cran version and not the latest github commit@main) and updated the README & DESCRIPTION to work with the actions / add badges.

For the pkgdown page you only have to set it to the gh-pages branch:
image
With the new manual action we can than choose which version of the package we want to build the site for.


kgoldfeld:
I still don't fully understand how the pkgdown is applied to the "correct" version.

PR #30, opened 2020-08-10
Add CoC, contribution guide etc.

assignUser:
The only thing "missing" for the repository now is a description with tags to make it easier to find the repo, only you as the owner can add that. Something like this:
image

and also upgraded cach@v1 to v2


kgoldfeld:
How about:

simstudy: Data generation to illuminate research

possible tags: r data-simulation data-generating-mechanisms

I am not sure what link we would include, since there is no official simstudy website: it could be https://www.rdatagen.net/page/simstudy/ or the new info create by pkgdown.

assignUser:
I like the tagline :D  for tags I would add data-generation.
As a link I would include the pgkdown site under https://kgoldfeld.github.io/simstudy once you set the source branch to gh-pages.
You will have to merge this pr as I can not approve my on pr ^^

kgoldfeld:
This "About" file - where does it go? Is there an example on dplyr?

And you said "once you set the source branch to gh-pages" - is that this PR?

PR #37, opened 2020-09-15
Restructure & integrate new formula system

assignUser:
As we are ready for the new version except for some minor bugfixes and additional documentation merging the large changes into main now is sensible. We can than use shorter lived branches to add the final touches.


assignUser:
\document After the merge I will run the rhub checks to see if there are any glaring issues.

PR #47, opened 2020-09-23
Unify genCorOrdCat/genOrdCat

### LINKED ISSUE #34: Unify genOrdCat & genCorOrdCat
~~Currently genOrdCat requires a variable to be specified in argument adjVar - there is no real reason this needs to be the case. If the value of adjVar is NULL, then the adjustment should be assumed to be 0 in all cases.~~

Merge genCorOrdCat into genOrdCat (with no correlation as default) allow creation 1 or many new cat vars. Deprecate genCorOrdCa (e.g. like [this](genCorOrdCa))

assignUser:
Closes #34


assignUser:
Need to check out the faile RCMD  when I have time.

assignUser:
This should be ready to merge now. @kgoldfeld have a look and merge if everything looks fine to you :)

assignUser:
After merging this we can update the vignette branche and check them with the new genOrdCat function, I already modified ordinal.Rmd on that branch.

kgoldfeld:
I know this has been merged - but in the future, we should be checking for valid corstr - (which are current "ind", "cs", and "ar1"). Right now, if it isn't one of those, it defaults to "ind", which is fine, but there should be a warning.

PR #53, opened 2020-09-25
Update Vignettes

### LINKED ISSUE #44: Vignettes
I just uploaded a new version of the "Overview" vignette to the Vignette branch. We need to see if it can be generated with the package. Tell me what I need to do in order to facilitate that - such as generate the html file. And, I want to make sure that it passes all the rhub checks, because that was giving me a problem a couple years ago, and that is why I moved everything to my website.

If this one works, I can move all the other vignettes over as well, and improve on them, of course, over time.

And if you think the overview is still not good, let me know. One thing I considered was including R examples for each of the distributions, but thought that might be too much. But, it would be easy to add. And any other suggestions  would be welcome.

assignUser:
Closes #44


kgoldfeld:
Is this something I should be merging?

kgoldfeld:
Should I be concerned with these errors?

assignUser:
@kgoldfeld  in a moment there was a Note about "scales" not being declared in suggests. And when fixing that I broke the description with an empty line :D so nothing serious

assignUser:
You can merge it now 👍

kgoldfeld:
I can merge even though there is still an error - I assume that will be fixed later on?

assignUser:
For unkown reasons on linux the hedgehog tests sometimes fail (CStack limit) but I think that is due to system limitations of the Actions runner. As the tests are skipped on cran anyway it is not of relevance for that. I will look into it at some point but it is rather low priority.

PR #54, opened 2020-09-25
Fix actions

assignUser:
The one issue with stac ksize should be solved on linux now, maybe we need to increase it more but I ran it a few times and it looks fine!

The other issue was random data generation during tests, most of the time they passed but if the seed was just wrong enough it fails. I will now set a custom seed on all tests that use random data with tolerances (e.g. genOrdCat).

You can merge this after the tests pass.

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -35,14 +35,6 @@ jobs:
 
       - uses: r-lib/actions/setup-pandoc@master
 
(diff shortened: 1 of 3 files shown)


assignUser:
Hmpf, I'll have another look.

assignUser:
So it looks like this finally fixed it. I needed to move the command to increase stack size into the same "run" block as otherwise a new process would be used (which would have the old, to small C stack size). 
Please merge when convenient.

Issue #43, opened 2020-09-19
check for negative entries in .adjustProbs


assignUser:
currently there is no check for negative probabilities in .adjustProbs

Issue #55, opened 2020-09-29
addCondition check for condition defintion fails incorrectly if definition is part of a list


kgoldfeld:
`addCondition` takes a "condition definition" as its first argument, which is a data.table of conditional definitions, like this:

```r
defS <- defCondition(defS, condition = "ss==2",  
                       formula = "(0.14 + a) * (C_rv==1) + (0.15 + a) * (C_rv==2) + (0.16 + a) * (C_rv==3)", 
                       dist = "nonrandom")
defS <- defCondition(defS, condition = "ss==3",  
                       formula = "(0.19 + a) * (C_rv==1) + (0.20 + a) * (C_rv==2) + (0.21 + a) * (C_rv==3)", 
                       dist = "nonrandom")

addCondition(defS, ...)
```

addCondidtion checks to see if defS exists, which in this case it does and everything works fine.

However, if we have created several definitions and included them in a list of definitions, so that `def_list <- list(defA = defA, defB = defB, defS = defS)`, this should work:

```r
addCondition(deflist[["defS"]]
```

but the check fails given the way it is set up. 

`addColumns` under this scenario works fine, and I suspect that is because the check is part of the robust new checking system. `addCondition` looks like it is still under the old system, and it fails.

Issue #51, opened 2020-09-25
Add genMixFormula


assignUser:
I had another idea for a useful function, particularly if you implment the vectorization. It could be nice to have a function `genMixFormula(functions, probs)` that takes two arguments - a vector of function names or a double-dot variable that is itself a vector and a vector of probabilities (which could be NULL, defaulting to equal probability *1/n* with *n* functions being mixed. 

_Originally posted by @kgoldfeld in https://github.com/kgoldfeld/simstudy/issues/44#issuecomment-698881921_

assignUser:
Do you mean something like this: 
```r
genMixFormula <- function(vars, probs = NULL, roundTo = 3) {
    assertNotMissing(vars = missing(vars))

    if (is.null(probs)) {
        n <- length(vars)
        probs <- round(rep(1 / n, n), roundTo)
    } else {
        assertNumeric(probs = probs)
        probs <- .adjustProbs(unlist(probs))
        assertLengthEqual(vars = vars, probs = probs)
    }

    paste(vars, probs, sep = " | ", collapse = " + ")
}

genMixFormula(c("a", "..b[..i]", "c"))
# [1] "a | 0.333 + ..b[..i] | 0.333 + c | 0.333"
genMixFormula(c("a", "..b", "c"), list(.2, .5, .3))
# [1] "a | 0.2 + ..b | 0.5 + c | 0.3"
```

I wasn't quite sure what you meant with
>or a double-dot variable that is itself a vector

kgoldfeld:
So, if `a` is a vector of values with length `n` and `b` is vector probabilities (also length `n`) in double-dot world, then I was thinking something like

```r
a <- c(1, 2, 5)
b <- c(.3, .2, .5)

genMixFormula(a, b}
# [1] "1 | .3 + 2 | .2  + 5 | .5"
```

Or it could be more general if we want to define the formula first and later on we are changing a and b:

```r
genMixFormula(a, b}
# [1] "..a[1] | ..b[1] + ..a[2] | ..b[2 ] + ..a[3] | ..b[3]"
```

But now that I think about it, I am not sure the second option makes sense, which is what I guess you are questioning.

I did have another thought - maybe it makes sense to have a single `genFormula()` function that can augment the existing function, incorporates `catProbs()`, and include as mixture" formula like you've started to create. Does this seem like a good idea, or simpler to keep as 3 separate functions?

assignUser:
[...]

PR #56, opened 2020-10-02
Fix a number of issues and add features

assignUser:
Fix issues #43 #51 #55 
also adds the shorthand to genMixFormula

Merge if everyhing passes and you thin its good 👍


assignUser:
👍

Issue #46, opened 2020-09-22
Another double-dot feature


kgoldfeld:
I am not adding this so that we necessarily think about whether or not it makes before the release of 0.2.0 (maybe this is for 0.3.0). (In general, do we have a list of things that we must complete before 0.2.0 is ready - as opposed to things that might be nice to include?)

The new idea would be to allow for vectors to be included in double-dot notation. Here is a non-vectorized case where it might be nice:

```r
library(simstudy)

size1 <- 2
size2 <- 4

def <- defData(varname = "blksize", 
    formula = "..size1 | .5 + ..size2 | .5", dist = "mixture")
genData(10, def)
#>     id blksize
#>  1:  1       4
#>  2:  2       4
#>  3:  3       4
#>  4:  4       2
#>  5:  5       2
#>  6:  6       2
#>  7:  7       4
#>  8:  8       4
#>  9:  9       4
#> 10: 10       2
```

However, it would be more compact to write as a vector, but obviously, it doesn't work:

```r
sizes <- c(2, 4)

def <- defData(varname = "blksize", 
               formula = "..sizes[1] | .5 + ..sizes[2] | .5", dist = "mixture")
#> Error in .checkMixture(newform): Invalid variable(s): 
#> Probabilities can only be numeric or numeric ..vars (not arrays). See ?distribution
```

assignUser:
There was (probably xD) a reason why I excluded this case and created an error but I can't remember why... I will check it out and test it with the check disabled.

kgoldfeld:
So, I don't mean to keep piling things on before we roll out 0.2.0. Is there a checklist of things we ***must*** do before releasing 0.2.0?

assignUser:
I have created a Milestone for the release which helps to keep track of the different issues. https://github.com/kgoldfeld/simstudy/milestone/1

assignUser:
I added what I think would make sense to do before the release

kgoldfeld:
Perfect. 

I have a general (unrelated) question for you. I noticed in the package help there are now descriptions of the distributions (and some other things that are not functions). Where is the source text for those descriptions?

assignUser:
https://github.com/kgoldfeld/simstudy/blob/main/R/simstudy-package.R

kgoldfeld:
Ah - thanks. 

[...]

Issue #41, opened 2020-09-18
double-dot notation for formula


kgoldfeld:
The double-dot notation for external variable reference is awesome. A couple of things. 

(1) In my workflow, I often put the data definitions at the top, and it seems natural to sometimes first reference the external below

```r
 #--- this doesn't work but is the way I like to do things

def <- defData(varname = "age", formula=10, dist = "nonrandom")
def <- defData(def, varname="agemult", 
               formula="age * ..age_effect", dist="nonrandom")
#> Error: Escaped variables referenced not defined (or not numeric): age_effect

age_effect <- 3
genData(2, def)
#>    id age
#> 1:  1  10
#> 2:  2  10

  #--- this does work as I am sure you know

age_effect <- 3

def <- defData(varname = "age", formula=10, dist = "nonrandom")
def <- defData(def, varname="agemult", 
               formula="age * ..age_effect", dist="nonrandom")

genData(2, def)
#>    id age agemult
#> 1:  1  10      30
#> 2:  2  10      30
```

(2)  The strength of the double-dot notation is how it makes dynamic data definition so easy. To generate different data sets in R using different assumptions, there are at least two ways. The first approach, using for loops, works fine, but I find it a little clunky and a bit less amenable to parallelization. (And, by the way, this is the perfect kind of case where it doesn't make sense to introduce `age_effect` before creating `def`. In fact, it is just not possible - except to create a dummy version of `age_effect`, which is not so appealing. But, maybe there's no way around this.

```r
  #--- this works but is a little clunky

list_of_data <- list()

age_effects <- c(0, 5, 10)
for (i in seq_along(age_effects)) {
  age_effect <- age_effects[i]
  list_of_data[[i]] <- genData(2, def)  
}

list_of_data
#> [[1]]
#>    id age agemult
#> 1:  1  10       0
#> 2:  2  10       0
#> 
#> [[2]]
#>    id age agemult
#> 1:  1  10      50
#> 2:  2  10      50
#> 
#> [[3]]
#>    id age agemult
#> 1:  1  10     100
#> 2:  2  10     100
```

I prefer to use lapply or mclapply - seems more compact and faster than the for loop, especially when doing 1000's of replications. But it doesn't work - and not sure how how hard it would be to make it work. 

```r
  #--- this doesn't work but is how I like to do things

myreplicate <- function(x, def) {
  age_effect <- x
  genData(2, def)
}

lapply(c(0, 5, 10), function(x) myreplicate(x, def))
[...]

PR #58, opened 2020-10-02
Move ..var checks to pre-generation

assignUser:
@kgoldfeld I think this should make the stuff possible you want to do in #41 Please check it out and let me know if this works for you or if you have any additional suggestions (mcapply will likely still not work) 

also renamed and deprecated catProbs


assignUser:
Also arrays now work as double dot vars, closes #46

Issue #59, opened 2020-10-03
make double dot work with mcapply


assignUser:
as mentioned in #41 thie does not work but should:
```r
  #--- this doesn't work but is how I like to do things

myreplicate <- function(x, def) {
  age_effect <- x
  genData(2, def)
}

lapply(c(0, 5, 10), function(x) myreplicate(x, def))
#> [[1]]
#>    id age agemult
#> 1:  1  10     100
#> 2:  2  10     100
#> 
#> [[2]]
#>    id age agemult
#> 1:  1  10     100
#> 2:  2  10     100
#> 
#> [[3]]
#>    id age agemult
#> 1:  1  10     100
#> 2:  2  10     100
```

assignUser:
@kgoldfeld this has to do with environments, I will look into how to maybe find way to solve this differntly but for now you can just use `<<-` for the assignment and it works:
``` r
library(simstudy)
def <- defData(varname = "age", formula=10, dist = "nonrandom")
def <- defData(def,
    varname = "agemult",
    formula = "age * ..age_effect", dist = "nonrandom"
)

myreplicate <- function(x, def) {
  age_effect <<- x
  genData(2, def)
}

lapply(c(0, 5, 10), function(x) myreplicate(x, def))
#> [[1]]
#>    id age agemult
#> 1:  1  10       0
#> 2:  2  10       0
#> 
#> [[2]]
#>    id age agemult
#> 1:  1  10      50
#> 2:  2  10      50
#> 
#> [[3]]
#>    id age agemult
#> 1:  1  10     100
#> 2:  2  10     100
```

<sup>Created on 2020-10-03 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0.9001)</sup>

assignUser:
maybe explaining this trick for simulations would be somethign for the double dot vignette #52

PR #60, opened 2020-10-03
Manage double-dot environments

assignUser:
So the issue with #59 and with the errors in the documentation I mentioned somewhere was that previously we used parent.frame() WITHIN simstudy. Due to the way environments work, if the ..var was not assigned in an environment upstream from simstudy e.g. the global environment we had no chance to finde the variable.

I add an `envir` argument to the functions that are used to generate data:
 - `addCorFlex`
 - `addColumns`
 - `addCondition`
 - `genData`
And of course all internal functions used by those. This allows for use of double-dot vars in mcapply and other situations with nested/different environments.

Let me know what you think!  

(It was easier then expected if this works correclty :D )

### DIFF
--- a/.github/workflows/pkgdown.yaml
+++ b/.github/workflows/pkgdown.yaml
@@ -1,6 +1,8 @@
 on:
   push:
-    branches: release
+    branches: 
+      - main
+      - release
 
 name: pkgdown
 
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Type: Package
 Package: simstudy
 Title: Simulation of Study Data
-Version: 0.2.0
+Version: 0.1.16.9000
 Date: 2020-09-25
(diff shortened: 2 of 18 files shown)


assignUser:
@kgoldfeld I just found a very cool feature of pkgdown that allows to have a site for the released version: https://kgoldfeld.github.io/simstudy/ and one for the dev version: https://kgoldfeld.github.io/simstudy/dev/ 
which is automatically decided by checking version number e.g. 0.2.0 -> release 0.1.16.9000 -> dev.
I think this is pretty cool and would like to use it, what do you think?

kgoldfeld:
The double-dot functionality is really cool, and works within functions, and by extension with lapply and mclapply. I think I can draft the vignette now.

kgoldfeld:
And yes - let's definitely have the two versions of the website - one for release, and one for development.

PR #68, opened 2020-10-07
Use backports

assignUser:
We had an [error on CRAN](https://cran.r-project.org/web/checks/check_results_simstudy.html)on windows oldrel. 
This was my failuire I attributed the issues before only to the stringi problem (which was true in the most part)
But I did not see the issue with the R 3.6.3. 
In R 4.0 new options were introduced to `suppressMessages` which will of course cause an unused argument error when used in R < 4.0 (this is only an issue for people trying to run the tests)
This is fixed by using the `backports` package.

Another issue is caused by R < 4.0 considering matrices to be of class `matrix` AND `vector` 
This causes an issue with one of the assert functions converting a matrix into a matrix with one row. 
(This will affect normal users )

I am very sorry for these oversights. I have already added an old release build to our CI to catch issues like this earlier. I will also send this build to rhub to double check.

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -19,7 +19,9 @@ jobs:
       matrix:
         config:
           - {os: windows-latest, r: 'release'}
+          - {os: windows-latest, r: 'oldrel'}
           - {os: macOS-latest, r: 'release'}
+          - {os: macOS-latest, r: 'oldrel'}
           - {os: ubuntu-20.04, r: 'release', rspm: "https://packagemanager.rstudio.com/cran/__linux__/focal/latest"}
           - {os: ubuntu-20.04, r: 'devel', rspm: "https://packagemanager.rstudio.com/cran/__linux__/focal/latest"}
 
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Type: Package
 Package: simstudy
 Title: Simulation of Study Data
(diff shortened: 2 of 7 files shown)


assignUser:
Ok our CI checks all passed. I also checked it with rhub on windows oldrel and devel and they also passed. The issues should be fixed now. Please merge and resubmit. I will tag the release afterwards.

assignUser:
Our readme head two "simstudy" titles for some reason. I fixed that. Only thing left for you to do will be to update cran comments.

kgoldfeld:
Does this effect our recent submission to CRAN?

assignUser:
@kgoldfeld  the bugs yes. we should resubmit. BTW I just checked and found that we should raise the minum required r version to 3.3.0 as RcppArmadillo (a dependency for mvnfast) requires that. So simstudy can not be installed on r < 3.3.0.

If we resubmit to fix the bugs anyway we can also adjust that.

kgoldfeld:
That's fine. I just re-rendered the README.md - as I read again, I am not sure I needed to do that. But it looks OK. What Version number will this be?

Tell me when I should proceed.

assignUser:
I have already increased the version number to 2.1.0 You can directly update cran-comments and the DESCRIPTION (for r version >= 3.3.0) afterwards you can submit to CRAN again (as this is a breaking bug that we patched it should be fine to sub mit again so soon?)

assignUser:
oh and add the version req rais to news md maybe?

PR #81, opened 2020-10-28
Assign user/issue40

### LINKED ISSUE #40: specifying "id" field for a new data set
The current approach to providing a the name for the "id" field is not ideal. It can be done either in the first call to defData or if no definition table is provided in genData, then it can be done there. I would like to think about making this consistent.

```
def <- defData(varname = "x1", dist = "binary", formula = .5, id = "cid")

genData(5, id = "xid")
genData(5, def, id = "xid")
```

assignUser:
Fixes #40 
This will now issue a warning when a definition is supplied and the provided id param does not match the definition:
image

### DIFF
--- a/R/generate_data.R
+++ b/R/generate_data.R
@@ -62,11 +62,11 @@ genData <- function(n, dtDefs = NULL, id = "id", envir = parent.frame()) {
 
     if (id != attr(dtDefs, "id") && !missing(id)) {
       valueWarning(
-        names = "id",
+        names = id,
         msg = list(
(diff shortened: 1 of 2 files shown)


kgoldfeld:
Can we make the warning a tad more clear? Something like:

The 'id' parameter is not valid when previously defined in call to function defData. This current specification is ignored.

assignUser:
Like this:
image

kgoldfeld:
Perfect - will merge after all checks are passed

PR #84, opened 2020-10-30
fix genOrdCat error when length(adjVar) == 1 & nCats >1

### LINKED ISSUE #83: genOrdCat with two responses but single adjustment variable generates error
The following snippet generates an error. In the previous multivariate version, do you recall if there was the ability to shift the probabilities using the adjustment variable? I am guessing not. If not, maybe we should consider fixing this, and if it was, then maybe we should definitely fix.

```r
library(simstudy)

d1 <- defData(varname = "z1", formula = 0, variance = .25)

dd <- genData(10, d1)

baseprobs <- matrix(c(.4, .3, .2, .1,
                      .5, .4, .05, .05), byrow = T, nrow = 2)

dd <- genOrdCat(dtName = dd, adjVar = "z1", baseprobs = baseprobs)
#> [1] 2
#> Error in rep(var, n): invalid 'times' argument
```

assignUser:
Fixes #83

The issue was caused when checking that adjVar is either of length 1 or same length as the number of categories.
For some reason I passed in a vector instead of a single number, not sure who hit me over the head on that one...

Should be fixed now. Sorry again.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,10 @@
 # simstudy (development version)
 * Added CITATION
-
+* genData now warns that it will ignore an 'id' parameter in case a definition
+  table is also passed.
+* Fix an error in genOrdCat when only a single adjustment variable is given but
+  more than one new category will be created.
+  
 # simstudy 0.2.2
 * Improve documentation and vignettes.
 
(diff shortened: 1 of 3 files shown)


assignUser:
The failed test were caused by the new test I inserted which modified the seed I previously set. Moved the test to the bottom to avoid that. I will have to spend some time on the tests anyway...

PR #114, opened 2021-11-02
make trtAssign/Obesere available as distributions

### LINKED ISSUE #69: Add trtAssignment as a "distribution" to defData to ease flow?
Currently, the treatment assignment process using `trtAssign` breaks the flow of the creation of a data set. Usually there is an outcome variable that is a function of the treatment assignment - so that we need to add a column to the table after the treatment assignment is made.

```r
# Data definitions (requires two definitions for the same data set)

d1 <- defData(varname = "g", formula = ".2;.4;.4", dist = "categorical")
d1 <- defData(d1, varname = "x1", formula = "0;10", dist = "uniform")
d1 <- defData(d1, varname = "x2", formula = "2+.5*2", variance = 3, dist = "normal")

d2 <- defDataAdd(varname = "y", formula = "2 + x2 * 3 + rx*2", variance = 5, dist = "normal")

# Data generation (requires three function calls)

dd <- genData(500, d1)
dd <- trtAssign(dd, strata = "g", grpName = "rx")
dd <- addColumns(d2, dd)
```

What if we added a "trtAssign" distribution to the data def table so that the treatment assignment can be part of a single data generation process? It would look like this:

```r
# Only one data definition

d <- defData(varname = "g", formula = ".2;.4;.4", dist = "categorical")
d <- defData(d, varname = "x1", formula = "0;10", dist = "uniform")
d <- defData(d, varname = "x2", formula = "2+.5*2", variance = 3, dist = "normal")
d <- defData(d, varname = "rx", formula = "1;1", variance = "g", dist = "trtAssign")
d <- defData(d, varname = "y", formula = "2 + x2 * 3 + rx*2", variance = 5, dist = "normal")

# Only one function call to generate the data

dd <- genData(500, d)
```
The formula for trtAssign represents the treatment assignment ratio defaults to "1;1", but could be of any length - so, if it is "1;1;1;2" that would be four groups. The variance parameter represents the stratification. Multiple levels of stratification would be represented as "a;b;c", where a, b, and c are variable names (really need to be categorical or factors). The functionality is exactly as it is in function `trtAssign`.

assignUser:
Closes #69 

I added both `trtAssing` and `trtObserve` as distributions. @kgoldfeld could you check it out and let me know if this works for you? I took the parameters assignment out of #69:
 
trtAssign:
* grpName = varname
* strata = variance
* ratio = formula
* balanced = link ("identity" will turn to `TRUE` everything else to `FALSE`)
* nTrt = `length(ratio)`

trtObserve
* grpName = varname
* fomulas = formula
* logit.link = link ("identity" will turn to `FALSE` everything else to `TRUE`)

I haven't added them to the distribution table or anything, if you could add some documentation that would be great!

### DIFF
--- a/R/define_data.R
+++ b/R/define_data.R
@@ -568,8 +568,10 @@ defSurv <- function(dtDefs = NULL,
       uniform = ,
 
       uniformInt = .checkUniform(newform),
+      trtAssign =  .checkCategorical(newform),
+      trtObserve = .isValidArithmeticFormula(newform),
 
-      stop("Unkown distribution.")
+      stop("Unknown distribution.")
     )
 
     invisible(newvar)
--- a/R/define_data.R
+++ b/R/define_data.R
@@ -569,7 +569,6 @@ defSurv <- function(dtDefs = NULL,
 
       uniformInt = .checkUniform(newform),
       trtAssign =  .checkCategorical(newform),
-      trtObserve = .isValidArithmeticFormula(newform),
 
       stop("Unknown distribution.")
     )
--- a/R/generate_data.R
+++ b/R/generate_data.R
@@ -77,8 +77,9 @@ genData <- function(n, dtDefs = NULL, id = "id", envir = parent.frame()) {
       id <- oldId %||% id
     }
 
-    dfSimulate <- data.frame(c(1:n)) # initialize simulated data with ids
-    names(dfSimulate) <- id # name first column attribute "id"
+    dfSimulate <- data.table::data.table(x = 1:n)
+    data.table::setnames(dfSimulate, id)
+    data.table::setkeyv(dfSimulate, id)
     iter <- nrow(dtDefs) # generate a column of data for each row of dtDefs
 
     for (i in (1:iter)) {
--- a/R/generate_dist.R
+++ b/R/generate_dist.R
@@ -105,6 +105,19 @@
       dtSim = dfSim,
       envir = envir
     ),
+    trtAssign = .genAssign(
+      dtName = dfSim,
+      grpName = args$varname,
+      strata = args$variance,
+      ratio = args$formula,
+      balanced = args$link
+    ),
+    trtObserve = .genObserve(
+      dt = dfSim,
+      grpName = args$varname,
+      formulas = args$formula,
+      logit.link = args$link
+    ),
     uniform = .genunif(
       n = n,
       formula = args$formula,
@@ -557,4 +570,31 @@
   unifCont <- stats::runif(n, range$min, range$max + 1)
 
   return(as.integer(floor(unifCont)))
+}
+
+.genAssign <- function(dtName, balanced,
+                      strata, grpName, ratio) {
+  dtName <- as.data.table(dtName)
+  if(strata == 0) {
+    strata <- NULL
+  } else {
+    strata <- .splitFormula(strata)
+  }
+  if (balanced == "identity") {
+    balanced <- TRUE
+  }
+  
+  ratio <- as.numeric(.splitFormula(ratio))
+  nTrt <- length(ratio)
+  trtAssign(dtName, nTrt, balanced, strata, grpName, ratio)[, ..grpName]
+                      }
+
+.genObserve <- function(dt, formulas, logit.link, grpName) {
+  dt <- as.data.table(dt)
+  if (logit.link == "identity") {
+    logit.link <- FALSE
+  } else {
+    logit.link <- TRUE
+  }
+  trtObserve(dt, formulas, logit.link, grpName)[, ..grpName]
 }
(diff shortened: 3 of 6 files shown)


kgoldfeld:
This is a very cool start - and will really impact my life (and hopefully others). Thanks so much for doing this.

The basic functionality works well:

```r
library(simstudy)
set.seed(1)

d1 <- defData(varname = "x", formula = 0, variance = 1)
d1 <- defData(d1, varname = "rx", formula = "1;1;1", dist = "trtAssign")

dd <- genData(1002, d1)
dd[,table(rx)]
#> rx
#>   1   2   3 
#> 334 334 334
```

I, do, however, get an error when trying to generate unbalanced assignment - and it doesn't matter what text is specified in `link`.

```r
### error

d1 <- defData(varname = "x", formula = 0, variance = 1)
d1 <- defData(d1, varname = "rx", formula = "1;1;1", dist = "trtAssign", link = "log")

dd <- genData(1002, d1)
#> Error in `[.data.table`(dt, , .SD, .SDcols = key(dt)): .SDcols should be column numbers or names
```

When stratifying, the results are not balanced:

```{r}
d1 <- defData(varname = "x", formula = 0, variance = 1)
d1 <- defData(d1, varname = "b", formula = .3, dist = "binary")
d1 <- defData(d1, varname = "rx", formula = "1;1;1", variance = "b", dist = "trtAssign")

dd <- genData(1002, d1)
dd[,table(b, rx)]
#>    rx
#> b     1   2   3
#>   0 222 234 242
#>   1 111 101  92
```

Here is how it should look, based on using the *function* `trtAssign`:

```r
dd <- trtAssign(dd, nTrt = 3, strata = "b")
dd[, table(b, trtGrp)]
#>    trtGrp
#> b     1   2   3
#>   0 233 233 232
#>   1 101 102 101
```

Some additional thoughts:

As it is implemented now, `nTrt` is calculated. This works well, but there is no way to specify it directly. I guess one way would be to allow an integer or a vector in  `formula` - and if it is an integer it defaults to the number of treatment groups, and if it is a vector, it represents the ratio (and number of groups). Not sure if that makes sense or is even possible.

I am fine with the name *trtAssign* for the distribution, but would it be worth considering something like *assignment* or *treatment*?

Finally, I was always ambivalent about `trtObserve` since it can really just be generated in the standard `defData` process using a *binary* or *categorical* distribution. I decided that it was OK as a standalone, since it allowed users to be explicit about the data generation process for treatment assignment, but I think it makes less sense as its own distribution in `defData` since it duplicates existing functionality.

assignUser:
I fixed the error, removed trtObserve as a dist and switched over the whole data generation to use data.table consistently, before it was a mix of dt and df. Maybe we will even see some speedup :D

kgoldfeld:
It seems to work great - I checked one more scenario - stratifying by multiple factors. This generates an error:

```r
library(simstudy)
set.seed(1)

### stratifying by two variables generates an error

d1 <- defData(varname = "x", formula = 0, variance = 1)
d1 <- defData(d1, varname = "b1", formula = .3, dist = "binary")
d1 <- defData(d1, varname = "b2", formula = .7, dist = "binary")
d1 <- defData(d1, varname = "rx", formula = "1;1;1", variance = c("b1", "b2"), dist = "trtAssign")

dd <- genData(1002, d1)
#> Error: Group name has previously been defined in data table
```

Using the function `trtAssign` directly:

```r
d1 <- defData(varname = "x", formula = 0, variance = 1)
d1 <- defData(d1, varname = "b1", formula = .3, dist = "binary")
d1 <- defData(d1, varname = "b2", formula = .7, dist = "binary")

dd <- genData(1002, d1)
dd <- trtAssign(dd, 3, strata = c("b1", "b2"))

dd[, .N, keyby = .(b1, b2, trtGrp)]
#>     b1 b2 trtGrp   N
#>  1:  0  0      1  78
#>  2:  0  0      2  77
#>  3:  0  0      3  78
#>  4:  0  1      1 163
#>  5:  0  1      2 162
#>  6:  0  1      3 163
#>  7:  1  0      1  29
#>  8:  1  0      2  29
#>  9:  1  0      3  28
#> 10:  1  1      1  65
#> 11:  1  1      2  65
#> 12:  1  1      3  65
```

assignUser:
That error is caused by the way you formated the `variance` 
`d1 <- defData(d1, varname = "rx", formula = "1;1;1", variance = c("b1", "b2"), dist = "trtAssign")` 
it should be like this (vectors will be possible in the new version though)
`d1 <- defData(d1, varname = "rx", formula = "1;1;1", variance = "b1;b2", dist = "trtAssign")`
this works without an error

kgoldfeld:
Yes - it is working, and all seems to be working well. I will merge.

Issue #109, opened 2021-10-30
Release simstudy 0.3.0


assignUser:
Issues to fix prior to release:
* [x]  Fix CRAN NOTE #104 
* [x] #102 
* [x] ? #94 @kgoldfeld I will need your feedback here about what exactly the issue is.
* [x] #69 

Prepare for release:

* [x] Check [current CRAN check results](https://cran.rstudio.org/web/checks/check_results_simstudy.html)
* [x] Check for quickly implemented issues
* [x] [Polish NEWS](https://style.tidyverse.org/news.html#news-release)
* [x] `devtools::build_readme()`
* [x] `urlchecker::url_check()`
* [x] `devtools::check(remote = TRUE, manual = TRUE)`
* [x] `devtools::check_win_devel()`
* [x] `rhub::check_for_cran()`
* [x] `rhub::check(platform = 'ubuntu-rchk')`
* [x] `rhub::check_with_sanitizers()`
* [x] Update `cran-comments.md`
* [x] Review pkgdown reference index for, e.g., missing topics

Submit to CRAN:

* [x] `usethis::use_version('minor')`
* [x] `devtools::submit_cran()`
* [x] Approve email

Wait for CRAN...

* [x] Accepted :tada:
* [x] `usethis::use_github_release()`
* [x] `usethis::use_dev_version()`

assignUser:
@kgoldfeld let me know if  there are any other thing that are a must fix/add before the release, I will start on the known issues.

kgoldfeld:
I don't think so.

assignUser:
@kgoldfeld I forgot to add news items for #102  and #105, also should we add documentation somewhere in a vignette/function documentation?

kgoldfeld:
Yes - I can definitely add some documentation in a vignette. I'll think about where it best fits, and will write something up. Actually, I will also use this example as the announcement on rdatagen.net

kgoldfeld:
My apologies - I am confused as to how we are handling vignettes. What is the workflow?

assignUser:
Do you mean github wise? Just create a new branch off of main.
For the actual editing: just add/edit the `.Rmd` files in `/vignettes` and knit via rstudio or `devtools::build_vignettes()`.
As I have never really done much with vignettes I can otherwise only point you to this rhub guide: https://blog.rhub.io/2020/06/03/vignettes/#vignette-101

kgoldfeld:
[...]

PR #115, opened 2021-11-02
Editing vignette

assignUser:
I have adjusted the heading levels (## instead of ###) so {pkgdown} automatically generates content sidebars for the vignettes. I also update all urls using {urlchecker} and bumped up the version and polished the news.

I think we should merge this once we are ready to submit with updated cran comments etc. (that's why I made it a draft for now)

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Type: Package
 Package: simstudy
 Title: Simulation of Study Data
-Version: 0.2.2.9000
+Version: 0.3.0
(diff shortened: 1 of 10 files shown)


kgoldfeld:
I want to update the distribution documentation to include the new *trtAssign* distribution, but can't do that in this branch, since the code is not up to date. Should I go ahead and merge this?

assignUser:
Yeah I think that should be fine, everything is merged and implemented and after you have updated the documentation wen can continue with the checks in #109 !

PR #133, opened 2022-01-12
Update pkgdown.yaml

assignUser:
I forgot to install simstudy, so the vignett building fails... ready to merge


kgoldfeld:
Thanks for fixing.

PR #134, opened 2022-01-13
Def data enhancements

kgoldfeld:
@assignUser This Includes two new functions `defRepeat` and `defRepeatAdd`. Tests have been included.

### DIFF
--- a/man/defRepeat.Rd
+++ b/man/defRepeat.Rd
@@ -6,7 +6,7 @@
 \usage{
 defRepeat(
   dtDefs = NULL,
-  nvars,
+  nVars,
   prefix,
   formula,
   variance = 0,
@@ -18,7 +18,7 @@ defRepeat(
 \arguments{
 \item{dtDefs}{Definition data.table to be modified}
 
-\item{nvars}{Number of new variables to define}
+\item{nVars}{Number of new variables to define}
 
 \item{prefix}{Prefix (character) for new variables}
 
@@ -42,8 +42,10 @@ Add multiple (similar) rows to definitions table
 The possible data distributions are: `r paste0(.getDists(),collapse = ", ")`.
 }
 \examples{
-def <- defRepeat(nvars = 4, prefix = "g", formula = "1/3;1/3;1/3", 
-   variance = 0, dist = "categorical")
+def <- defRepeat(
+  nVars = 4, prefix = "g", formula = "1/3;1/3;1/3",
+  variance = 0, dist = "categorical"
+)
 def <- defData(def, varname = "a", formula = "1;1", dist = "trtAssign")
 def <- defRepeat(def, 8, "b", formula = "5 + a", variance = 3, dist = "normal")
 def <- defData(def, "y", formula = "0.10", dist = "binary")
--- a/man/defRepeatAdd.Rd
+++ b/man/defRepeatAdd.Rd
@@ -7,7 +7,7 @@ existing data.table}
 \usage{
 defRepeatAdd(
   dtDefs = NULL,
-  nvars,
+  nVars,
   prefix,
   formula,
   variance = 0,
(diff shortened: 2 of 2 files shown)


kgoldfeld:
@assignUser This was a tad premature - I haven't added anything to the vignettes - I need to do that.

kgoldfeld:
@assignUser Vignette is updated, so I think it is ready to go.

assignUser:
I'll check it out this weekend :)

R/define_data.R
```diff
@@ -216,6 +216,135 @@ defDataAdd <- function(dtDefs = NULL,
...
+    defNew <- data.table::copy(dtDefs)
+    
+    for (i in 1:nvars) {
+      defNew <- defData(defNew, varname = varnames[i], 
+                  formula = formula, variance = variance, 
+                  dist=dist, link = link, id = id)
+    }
+  }
+  
+  return(defNew[])
```
assignUser, line 280: As an aside why the []?

R/define_data.R
```diff
@@ -216,6 +216,135 @@ defDataAdd <- function(dtDefs = NULL,
   return(defNew[])
 }
 
+#' Add multiple (similar) rows to definitions table
+#'
+#' @param dtDefs Definition data.table to be modified
+#' @param nvars Number of new variables to define
```
assignUser, line 222: Ill rename this to nVars to be consistent with camelCase

R/define_data.R
```diff
@@ -216,6 +216,135 @@ defDataAdd <- function(dtDefs = NULL,
...
+#' @seealso [distributions]
+#' @details The possible data distributions are: `r paste0(.getDists(),collapse = ", ")`.
+#'
+#' @examples
+#' def <- defRepeatAdd(nvars = 4, prefix = "g", formula = "1/3;1/3;1/3", 
+#'     variance = 0, dist = "categorical")
+#' def <- defDataAdd(def, varname = "a", formula = "1;1", dist = "trtAssign")
+#' def <- defRepeatAdd(def, 8, "b", formula = "5 + a", variance = 3, dist = "normal")
+#' def <- defDataAdd(def, "y", formula = "0.10", dist = "binary")
+#'
```
assignUser, line 304: Do we need `defRepeatAdd` in addition to `defRepeat` as they do the same thing? Code duplication like this is usally a big no-no :D

assignUser:
\document

R/define_data.R
kgoldfeld: data.tables behave in strange ways sometimes, and providing the brackets ensures that the data.table being returned from a function is available immediately. This behavior may have changed, but there is no harm in including it. If you do choose to eliminate it, definitely make sure everything behaves properly.

R/define_data.R
kgoldfeld: Well - they do the same thing, except for the variable checking, just like `defData` and `defDataAdd`. `defRepeatAdd` calls `defDataAdd`, whereas `defRepeat` calls `defData`.

R/define_data.R
kgoldfeld: Sounds good.

R/define_data.R
assignUser: Hmm I see. Ah that remindes  me of #18 
But that can still be done without complete duplication of code... I`ll try it later

kgoldfeld: Alternatively, `defRepeat` wouldn't be needed at all if `defData` had another argument `nVars` that would just indicate how many instances of the current variable. But, I don't mind the extra function, since it makes it explicit what the user is trying to do. Again, complete transparency at the expense of efficiency.

kgoldfeld:
@assignUser Just wanted to confirm that we are on board with this, and I can go ahead and merge ...

assignUser:
Sure, go ahead we will deal with the code duplication when handeling #18  as it is basically the same thing :D

Issue #135, opened 2022-01-15
Categorical distribution fails in one very specific case


kgoldfeld:
`genData` fails to generate categorical when the formula is specified as `".1;.2;.4;.2;.1"`, but works fine with other possible configurations.

```r
d1 <- defData(varname = "x1", formula = ".1;.2;.4;.2;.1", dist = "categorical")
d1 <- defData(d1, varname = "x2", formula = ".1;.2;.4;.1;.2", dist = "categorical")
d1 <- defData(d1, varname = "x3", formula = ".1;.2;.5;.1;.1", dist = "categorical")
d1 <- defData(d1, varname = "x4", formula = ".1;.8;.1", dist = "categorical")

dx <- genData(1000, d1)
dx[, .N, keyby = x1]
#>    x1    N
#> 1:  0 1000
dx[, .N, keyby = x2]
#>    x2   N
#> 1:  1 111
#> 2:  2 196
#> 3:  3 409
#> 4:  4  79
#> 5:  5 205
dx[, .N, keyby = x3]
#>    x3   N
#> 1:  1 110
#> 2:  2 211
#> 3:  3 498
#> 4:  4  78
#> 5:  5 103
dx[, .N, keyby = x4]
#>    x4   N
#> 1:  1  91
#> 2:  2 817
#> 3:  3  92
```

assignUser:
Thats's weird, I'll have a look. You have more experience with the underlying stats functions, any idea what could cause this?

kgoldfeld:
I found it and fixed it. It was a strange rounding error that seems to only happen in that particular case. It took place in `.gencat`, and the added row had a value of -2e-16 when it should be 0. 

```r
parsedProbs <- cbind(parsedProbs, 1 - rowSums(parsedProbs))
```

I added a line to round all probability values to 12 decimal places to drive that number to 0. I can't guarantee that it will be fixed for all cases, but my guess is it will be.

PR #136, opened 2022-01-15
Fixing rounding bug in .gencat

kgoldfeld:
This is a fix for Issue #135.


assignUser:
Should we mention the rounding in the gencat documentation? 12 digits should be enough for most anyone but you never know...

assignUser:
@kgoldfeld the failed check you might have an email about was caused by the action that sets up the r dependencies, I think it is a timeout issue or something as I had it yesterday and re-running the check solved it.

kgoldfeld:
I will add documentation to make it clear, though it would be pretty useless to simulate data based on a probability of 1e-12. I just tested it out with `rbinom(1e8, 1, prob = 1e-12), and ran that 100 times, so we effectively had 1e10 (10 billion) tosses, and no successes. Maybe with a trillion tosses, we'll see one - at least that is what we expect. I am not sure `simstudy` can handle a trillion tosses.

assignUser:
Wow :grin:

kgoldfeld:
Any objections to my merging (assuming that last check passes)?

assignUser (approved):
Looks good to me!

PR #140, opened 2022-02-04
Add survival param utility

kgoldfeld:
This modification includes two new utility functions `survGetParams` and `survPlot` that help users identify parameters to use in function `genSurv`. Error checking has been added, as well as code for testing.


kgoldfeld:
@assignUser If you want to take a look at this and let me know if it looks good, that would be great. The will be a blog post coming out early next week describing the motivation for these new functions.

R/utility.R
```diff
@@ -725,4 +724,117 @@ viewSplines <- function(knots, degree, theta) {
...
+#' points <- list(c(60, 0.90), c(100, .75), c(200, .25), c(250, .10))
+#' survGetParams(points)
+#' @export
+#' @concept utility
+survGetParams <- function(points) {
+  
+  assertNotMissing(points = missing(points))
+  assertClass(points = points, class = "list")
+  
+  time <- sapply(points, function(x) x[1])
```
assignUser, line 746: As described [here](https://adv-r.hadley.nz/functionals.html?q=sapply#purrr-shortcuts) (scroll up a bit from the linked section) it is safer to use vapply then sapply. `FUN.VALUE` describes a template of the return value of the function, here simply a numeric of length 1. and will error on anything else. sapply might try other simplifications that could cause issues. 
```suggestion
  time <- vapply(points, function(x) x[1], FUN.VALUE = numeric(1))
```

R/utility.R
```diff
@@ -725,4 +724,117 @@ viewSplines <- function(knots, degree, theta) {
...
+  p <- sapply(points, function(x) x[2])
```
assignUser, line 747:
```suggestion
  p <- vapply(points, function(x) x[2],  FUN.VALUE = numeric(1))
```

R/utility.R
```diff
@@ -725,4 +724,117 @@ viewSplines <- function(knots, degree, theta) {
...
+  assertProbability(p)  
+  assertDescending(p)
+  
+  loss_surv <- function(params, points) {
+    
+    loss <- function(a, params) {
+      ( params[2]*(log(-log(a[2])) - params[1]) - log(a[1]) ) ^ 2
+    }
+    
+    sum(sapply(points, function(a) loss(a, params)))
```
assignUser, line 761:
```suggestion
    sum(vapply(points, function(a) loss(a, params), FUN.VALUE = numeric(1)))
```

R/utility.R
```diff
@@ -725,4 +724,117 @@ viewSplines <- function(knots, degree, theta) {
...
+    
+  }
+  
+  r <- stats::optim(
```
assignUser, line 765: It is just a style thing but one letter variable names should be avoided in most cases for readability, (I know statisticians love them :grin: ). As this is only used in two pretty unambiguous places it is not critical but worth mentioning.

R/utility.R
```diff
@@ -725,4 +724,117 @@ viewSplines <- function(knots, degree, theta) {
...
+  
+  if (r$convergence !=0) stop("Optimization did not converge")
+  
+  return(r$par)
+}
+
+#' Plot survival curves
+#' 
+#' @param f This is the "formula" parameter of the Weibull-based survival curve
+#' that can be used to define the scale of the distribution
```
assignUser, line 782:
```suggestion
#' that can be used to define the scale of the distribution.
```

R/utility.R
```diff
@@ -725,4 +724,117 @@ viewSplines <- function(knots, degree, theta) {
...
+  
+  dd <- data.table::data.table(
+    f = f,
+    scale = scale,
+    shape = shape,
+    T = (-(log(u)*scale/exp(f)))^(shape),
+    p = round(1 - cumsum(rep(1/length(u), length(u))), 3)
+  )
+  
+  p <- ggplot2::ggplot(data = dd, ggplot2::aes(x = T, y = p)) +
```
assignUser, line 822: As we only have `ggplot2` in `SUGGESTS` we would either need to add it as an `IMPORT` or guard against the usage (which I think is the preferred option as we only use it in 3 functions that are very non-essential) as suggested [here](https://r-pkgs.org/description.html#guarding-the-use-of-a-suggested-package). 

I just checked and `ViewSplines` and `ViewBasis` should also get the same treatment.

R/utility.R
```diff
@@ -725,4 +724,117 @@ viewSplines <- function(knots, degree, theta) {
   }
 
   return(p)
-}
\ No newline at end of file
+}
+
+#' Get survival curve parameters
+#'
+#' @param points A list of two-element vectors specifying the desired time and
```
assignUser, line 731: Is there a usability reason why you require a (for R rather unusual) list of two element vectors instead of a more "natural"  data.frame or matrix? (same for the plot function ofc)

R/utility.R
```diff
@@ -725,4 +724,117 @@ viewSplines <- function(knots, degree, theta) {
...
+  )
+  
+  if (r$convergence !=0) stop("Optimization did not converge")
+  
+  return(r$par)
+}
+
+#' Plot survival curves
+#' 
+#' @param f This is the "formula" parameter of the Weibull-based survival curve
```
assignUser, line 781: Here maybe `weibull_formula` or `formula`  would be a bit  more descriptive (unless ofc the parameter is actually called f in the underlying distribution).

assignUser (changes requested):
A few comments from me but nothing really substantial, great work overall and really cool that you are getting into testing!

R/utility.R
kgoldfeld: Done.

R/utility.R
kgoldfeld: Changed to `optim_results`.

R/utility.R
kgoldfeld: Changed to `formula`

R/utility.R
kgoldfeld: I did go back and forth on this, but I kind of liked the explicit nature of entering in the pair together. Since this is just a utility function that is really meant to be used on the fly (though it could be part of reproducible code as well), it seemed to make sense. Also, in most cases, folks will only be entering two points, maybe three, so it seemed easy enough. I don't mind being a little different. So, for now, let's keep as is.

assignUser (approved):

PR #150, opened 2022-05-04
defDummy tests

elinsooon:


kgoldfeld (commented):
Nice - I'll test it out later, but it looks good from what I see. 

To start to answer your question about factors and integers, integers are numeric values that take on only integer (whole number) values: ..., -2, -1, 0, 1, 2, 3, .... Factors are another data type and is the way that R deals with categorical data. So, we might have three temperature categories "Hot", "Moderate", and "Cold" - so those could be stored as just characters or factors. If they are stored as characters, they really have no meaning at all. But as factors, they can be analyzed and summarized in powerful ways. (Sometimes R will automatically convert a character variable to a factor if you try to analyze it.) The dummy function is designed to work specifically with factors. 

Might be good to generate data using the genDummy and take a look at the data generated - if you haven't already. Always good to see it in action. And I am happy to explain what it is doing in case it makes no sense. 

[Here](https://swcarpentry.github.io/r-novice-inflammation/12-supp-factors/index.html) is a general description of factors.

tests/testthat/test-generate_data.R
```diff
@@ -280,3 +280,38 @@ test_that("genFactor works.", {
...
+  set.seed(10076)
+  
+  d1 <- defData(varname = "rx", formula = "1;1", dist = "trtAssign")
+  d1 <- defData(d1, varname = "male", formula = .4, dist = "binary")
+  d1 <- defData(d1, varname = "z", formula = "0 - 1.2*rx - 1*male", dist = "nonrandom")
+  
+  dd <- genData(5, d1)
+  
+  # Initial data checks
+  expect_error(genDummy(varname = "rx", sep = ".", replace = FALSE))
```
assignUser, line 296: [`expect_error`](https://testthat.r-lib.org/reference/expect_error.html#arguments) has some features that allow you to test for specific errors, either by regex matching against the error message or by checking if the error is of the correct class e.g. [simstudy::missingArgument](https://github.com/kgoldfeld/simstudy/blob/main/R/conditions.R#L29), this is important as it prevents the test from passing in case there is some other error you were not expecting.

tests/testthat/test-generate_data.R
```diff
@@ -280,3 +280,38 @@ test_that("genFactor works.", {
...
+  expect_error(genDummy(dd, sep = ".", replace = FALSE))
+  
+  # Check if data table exists
+  expect_error(genDummy(dt, varname = "rx", sep = ".", replace = FALSE))
+  
+  # Check if varname exists
+  expect_error(genDummy(dd, varname = "xx", sep = ".", replace = FALSE))
+  
+  #I do not understand how to test for is.integer or is.factor, I'm not clear on the
```
assignUser, line 305: In addition to what Keith already said, the best way to test for them is [expect_type](https://testthat.r-lib.org/reference/inheritance-expectations.html).

assignUser: Also the [adv-r book](https://adv-r.hadley.nz/vectors-chap.html#factors) has some info on this, overall it and [r-pkgs](https://r-pkgs.org/) are great resources :)

assignUser:
👍 I just added a bit off information but good work so far! 🚀  
I am a bit hard pressed on time at the moment but please don't hesitate to ping me for any technical questions you might need input on!

elinsooon:
I'll look into integers and factors a bit more. I thought I had seen something in the documentation that threw me for a loop with integers, but I might've read it wrong since that is pretty straightforward.

I had tried to implement the error classes initially but with the way the function is written, it only returns a written error label, not an actual error class, I'll take a look at expect_type.

Thanks for the info!

tests/testthat/test-generate_data.R
```diff
@@ -280,3 +280,38 @@ test_that("genFactor works.", {
...
+  d1 <- defData(d1, varname = "z", formula = "0 - 1.2*rx - 1*male", dist = "nonrandom")
+  
+  dd <- genData(5, d1)
+  
+  # Initial data checks
+  expect_error(genDummy(varname = "rx", sep = ".", replace = FALSE))
+  expect_error(genDummy(dd, sep = ".", replace = FALSE))
+  
+  # Check if data table exists
+  expect_error(genDummy(dt, varname = "rx", sep = ".", replace = FALSE))
```
kgoldfeld, line 300: Just a warning - "dt" is actually a function name in base R (relating to the student-t density). It is possible to create a variable with that name - and I used to do that. But, it isn't great practice. In any case, probably best not to use that anywhere except when referring to the function itself. "DT" is a fine alternative, or anything else that is not a function.

kgoldfeld:
Closing now without merging - as it was replaced.

PR #152, opened 2022-05-06
genDummy tests and other small optimizations

elinsooon:
genDummy code optimizations
trtAssign optimizations
genDummy tests
some grammar fixes in comments


max_test.Rmd
```diff
@@ -59,6 +59,8 @@ defNew
 
 ```
 ```{r}
-assertInteger(as.integer(1))
+test <- c("one", "two", "three")
+
+typeof(assertIntegerOrFactor(var1 = test))
 ```
```
kgoldfeld, line 66: This test file should probably not be saved in the package, but in another workspace.

R/asserts.R
```diff
@@ -148,6 +148,52 @@ assertInteger <- function(..., type, call = sys.call(-1)) {
     }
 }
 
+#' Check for Factor
+#'
+#' @description Checks if all passed vars and their content are factors or can
+#' be coerced to factor without loss of information.
+#' @param ... Any number of variables as named elements e.g. var1 = var1.
+#' @noRd
+assertFactor <- function(..., type, call = sys.call(-1)) {
```
kgoldfeld, line 157: We will also need tests for these new assert functions

max_test.Rmd
elinsooon: i just moved it

R/asserts.R
elinsooon: I didn't end up using it so I just commented it out, I didn't want to delete it outright but let me know if I should store it somewhere besides right inside the code

R/generate_dist.R
```diff
@@ -130,13 +130,17 @@
     )
   )
 
+  if (args$dist == "trtAssign") {
+    newColumn <- as.integer(newColumn)
+  }
+
```
kgoldfeld, line 136: I think it would be better to modify the `trtAssign` function [directly](https://github.com/kgoldfeld/simstudy/blob/403a6adadf6bb34bf396a68b4ab622a5c3fcd1b2/R/group_data.R#L354), rather than including it here. That way if users generate treatment assignments more generally (it can be done outside of `defData`) will also be constrained to integer results.

R/generate_data.R
```diff
@@ -137,28 +137,29 @@ genData <- function(n, dtDefs = NULL, id = "id", envir = parent.frame()) {
...
   # Initial data checks
-
-  if (missing(dtName)) stop("argument 'dtName' is missing", call. = FALSE)
-  if (missing(varname)) stop("argument 'varname' is missing", call. = FALSE)
+  
+  assertNotMissing(dtName = missing(dtName), varname = missing(varname))
 
   # Check if data table exists
 
   if (!exists(deparse(substitute(dtName)), envir = parent.frame())) {
```
kgoldfeld, line 145: Maybe we should create `assertDataTableExists` function. Another way we've checked the data table ([addColumns](https://github.com/kgoldfeld/simstudy/blob/403a6adadf6bb34bf396a68b4ab622a5c3fcd1b2/R/add_data.R#L27), for example) is by using `assertClass`, but this doesn't seem quite right, because it doesn't give the precise reason for error failure.

R/asserts.R
kgoldfeld: Was there a particular reason you didn't implement the general assert functions. I think we should go ahead and implement them.

kgoldfeld: Yeah - I think it would be worthwhile - as it will possibly avoid duplication of code down the road (either for new functionality or for cleaning up existing code).

R/generate_data.R
elinsooon: That was my initial thought when making this, but I had some trouble trying to construct a function because if I tried to pass a dtName that didn't exist to the assert function, it would automatically return an error and wouldn't allow the assert to run. I'll take another look to see if I can figure something out.

R/generate_dist.R
elinsooon: Just updated this

R/asserts.R
elinsooon: I uncommented assertFactor and added a test_that. I'm having trouble coming up with a case that would fail the assert because as I understand it the only thing that stops something from being coerced into a factor is if there is an infinite amount of options in the vector, dt, etc that it's in. Unless assertFactor changes to be just whether a value is a factor, and not whether it can be coerced into one, I'm not sure that it would ever return an error.

R/generate_data.R
elinsooon: I haven't been able to solve this problem, and the workaround that I did sort of find doesn't seem to work either

tests/testthat/test-generate_data.R
```diff
@@ -280,3 +280,104 @@ test_that("genFactor works.", {
...
+    
+    if (dd4dum$a[i] == 3) {
+      expect_equal(dd5dum$a.1[i], 0)
+      expect_equal(dd5dum$a.2[i], 0)
+      expect_equal(dd5dum$a.3[i], 1)
+    }
+    
+  }
+  
+})
```
kgoldfeld, line 383: I like these tests to see if it works - nice.

R/asserts.R
kgoldfeld: Looks good now.

PR #155, opened 2022-05-12
genFormula optimization and testing, genMarkov work

elinsooon:
genFormula optimizations
genFormula test_that throws errors

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,8 @@
 ## New features
 * Updated genFormula with functionality to generate formulas with double dot variables inside.
 
+* Added argument 'startProb' to genMarkov. Allows user to set probability distribution of start state.
+
 * Added utility functions survGetParams and survParamPlot to aid users in identifying parameters that can be used to generate desired distributions of time to event data.
 
 * Major updates to functions defSurv and genSurv. It is now possible to generate survival outcomes with hazard functions that change over time. In addition, competing risk outcomes can be explicitly generated.
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,8 @@
 # simstudy (development version)
 
 ## New features
+* Updated genFormula with functionality to generate formulas with double dot variables inside.
+
 * Added utility functions survGetParams and survParamPlot to aid users in identifying parameters that can be used to generate desired distributions of time to event data.
 
 * Major updates to functions defSurv and genSurv. It is now possible to generate survival outcomes with hazard functions that change over time. In addition, competing risk outcomes can be explicitly generated.
--- a/R/add_data.R
+++ b/R/add_data.R
@@ -260,29 +260,58 @@ addMarkov <- function(dd, transMat, chainLen, wide = FALSE, id = "id",
   # 'declare' vars created in data.table
   variable <- NULL
   .e <- NULL
-
-  # check transMat is square matrix and row sums = 1
-
-  if (!is.matrix(transMat) |
-    (length(dim(transMat)) != 2) |
-    (dim(transMat)[1] != dim(transMat)[2])
-  ) {
-    stop("Transition matrix needs to be a square matrix")
+  
+  ######
+  # check transMat is matrix
+  if (!is.matrix(transMat)) {
+    c <- condition(c("simstudy::typeMatrix", "error"),
+                   "transMat is not a matrix!")
+    stop(c)
   }
 
-  # check row sums = 1
+  # check transMat is square matrix
+  if ((length(dim(transMat)) != 2) |
+      (dim(transMat)[1] != dim(transMat)[2])) {
+    c <- condition(c("simstudy::squareMatrix", "error"),
+                   "transMat is not a square matrix!")
+    stop(c)
+  }
 
+  # check transMat row sums = 1
   if (!all(round(apply(transMat, 1, sum), 5) == 1)) {
-    stop("Rows in transition matrix must sum to 1")
+    c <- condition(c("simstudy::rowSums1", "error"),
+                   "transMat rows do not sum to 1!")
+    stop(c)
   }
 
-  # check chainLen is > 1
+  # check chainLen greater than 1
+  if (chainLen <= 1) {
+    c <- condition(c("simstudy::chainLen", "error"),
+                   "chainLen must be greater than 1!")
+    stop(c)
+  }
+  
+  # if start0lab defined, check that it is defined in dd
+  if (!is.null(start0lab)) {
+    assertInDataTable(vars = start0lab, dt = dd)
+    
+  }
 
-  if (chainLen <= 1) stop("Chain length must be greater than 1")
+  # if start0lab defined, check that it exists in the transition matrix
+  if (!is.null(start0lab)) {
+    if (any(0 < dd[, start0lab, with = FALSE] | dd[, start0lab, with = FALSE] > dim(transMat)[1])) {
+      c <- condition(c("simstudy::start0probNotInTransMat", "error"),
+                     "all start states in start0prob must exist in the transistion matrix!")
+      stop(c)
+    }
+
+  }
+  ######
 
   # verify id is in data.table dd
 
-  if (!(id %in% names(dd))) stop(paste(id, "is not in data table"))
+  #if (!(id %in% names(dd))) stop(paste(id, "is not in data table"))
+  assertInDataTable(vars = id, dt = dd)
 
   ####
 
@@ -290,8 +319,6 @@ addMarkov <- function(dd, transMat, chainLen, wide = FALSE, id = "id",
 
   if (is.null(start0lab)) {
     s0 <- rep(1, n)
-  } else if (!(start0lab %in% names(dd))) {
-    stop(paste("Start state field", start0lab, "does not exist"))
   } else {
     s0 <- dd[, get(start0lab)]
   }
--- a/R/add_data.R
+++ b/R/add_data.R
@@ -299,7 +299,7 @@ addMarkov <- function(dd, transMat, chainLen, wide = FALSE, id = "id",
 
   # if start0lab defined, check that it exists in the transition matrix
   if (!is.null(start0lab)) {
-    if (any(0 > dd[, start0lab, with = FALSE] | dd[, start0lab, with = FALSE] > dim(transMat)[1])) {
+    if (any(1 > dd[, start0lab, with = FALSE] | dd[, start0lab, with = FALSE] > dim(transMat)[1])) {
       c <- condition(c("simstudy::start0probNotInTransMat", "error"),
                      "all start states in start0prob must exist in the transistion matrix!")
       stop(c)
--- a/R/add_data.R
+++ b/R/add_data.R
@@ -299,7 +299,7 @@ addMarkov <- function(dd, transMat, chainLen, wide = FALSE, id = "id",
 
   # if start0lab defined, check that it exists in the transition matrix
   if (!is.null(start0lab)) {
-    if (any(0 < dd[, start0lab, with = FALSE] | dd[, start0lab, with = FALSE] > dim(transMat)[1])) {
+    if (any(0 > dd[, start0lab, with = FALSE] | dd[, start0lab, with = FALSE] > dim(transMat)[1])) {
       c <- condition(c("simstudy::start0probNotInTransMat", "error"),
                      "all start states in start0prob must exist in the transistion matrix!")
       stop(c)
--- a/R/asserts.R
+++ b/R/asserts.R
@@ -43,8 +43,7 @@ assertEqual <- function(..., val, call = sys.call(-1)) {
   
   notEqual <- !sapply(dots$args, function(i) {i == val})
   if (any(notEqual)) {
-    notEqualError(dots$names, val, call = call,
-                  msg = "{names *} should be equal to {val}")
+    notEqualError(dots$names, val, call = call)
   }
 }
 
--- a/R/asserts.R
+++ b/R/asserts.R
@@ -139,8 +139,8 @@ assertInteger <- function(..., type, call = sys.call(-1)) {
     assertNumeric(..., call = call)
     dots <- dots2argNames(...)
     checkInteger <- function(arg) {
-        arg <- unlist(arg)
-        all(arg == as.integer(arg))
+      arg <- unlist(arg)
+      all(arg == as.integer(arg))
     }
     notInteger <- !sapply(dots$args, checkInteger)
     if (any(notInteger)) {
@@ -150,18 +150,13 @@ assertInteger <- function(..., type, call = sys.call(-1)) {
 
 #' Check for Factor
 #'
-#' @description Checks if all passed vars and their content are factors or can
-#' be coerced to factor without loss of information.
+#' @description Checks if all passed vars and their content are factors.
 #' @param ... Any number of variables as named elements e.g. var1 = var1.
 #' @noRd
 assertFactor <- function(..., type, call = sys.call(-1)) {
   #assertNumeric(..., call = call)
   dots <- dots2argNames(...)
-  checkFactor <- function(arg) {
-    arg <- unlist(arg)
-    all(arg == as.factor(arg))
-  }
-  notFactor <- !sapply(dots$args, checkFactor)
+  notFactor <- !sapply(dots$args, function(fac) {is.factor(fac)})
   if (any(notFactor)) {
     typeError(dots$names[notFactor], type = "factor", call = call)
   }
@@ -169,30 +164,37 @@ assertFactor <- function(..., type, call = sys.call(-1)) {
 
 #' Check for Integer OR Factor
 #'
-#' @description Checks if all passed vars and their content are factors or can
-#' be coerced to factor without loss of information.
+#' @description Checks if all passed vars and their content are integers or factors.
 #' @param ... Any number of variables as named elements e.g. var1 = var1.
 #' @noRd
-#assertIntegerOrFactor <- function(..., type, call = sys.call(-1)) {
-#  #assertNumeric(..., call = call)
-#  dots <- dots2argNames(...)
-#  checkInteger <- function(arg) {
-#    arg <- unlist(arg)
-#    all(arg == as.integer(arg))
-#  }
-#  checkFactor <- function(arg) {
-#    arg <- unlist(arg)
-#    all(arg == as.factor(arg))
-#  }
-#  notInteger <- !sapply(dots$args, checkInteger)
-#  notFactor <- !sapply(dots$args, checkFactor)
-#  #notInteger <- assertInteger(...)
-#  #notFactor <- assertFactor(...)
-#  if (!notFactor && !notInteger) {
-#    typeError(dots$names[notIntegerOrFactor], type = type, call = call)
-#    #typeError(notIntegerOrFactor, type = type, call = call)
-#  }
-#}
+# assertIntegerOrFactor <- function(..., type, call = sys.call(-1)) {
+#   #assertNumeric(..., call = call)
+#   dots <- dots2argNames(...)
+#   #checkIntegerOrFactor <- function(arg) {
+#   #  arg <- unlist(arg)
+#   #  all(assertInteger(var1 = arg) | assertFactor(var1 = arg))
+#   #}
+#   #notInteger <- !sapply(dots$args, checkInteger)
+#   
+#   
+#   #notIntOrFac <- !sapply(dots$args, function(x) {((x == as.integer(x)) | is.factor(x))})
+# 
+#   checkInteger <- function(arg) {
+#     arg <- unlist(arg)
+#     all(tryCatch({assertNumeric(arg)},
+#                  error = FALSE,
+#                  finally = {
+#                    arg == as.integer(arg)
+#                  }))}
+# 
+#   
+#   notInteger <- !sapply(dots$args, checkInteger)
+#   notFactor <- !sapply(dots$args, function(fac) {is.factor(fac)})
+#   
+#   if (any(notInteger && notFactor)) {
+#     typeError(dots$names[notIntOrFac], type = "integer or factor", call = call)
+#   }
+# }
 
 #' Check for Value
 #'
@@ -238,12 +240,13 @@ assertUnique <- function(..., call = sys.call(-1)) {
 #' @description Checks if dt exists in environment.
 #' @param dtName Name of dt to check (as a string).
 #' @noRd
-#assertDataTableExists <- function(dtNameStr, call = sys.call(-1)) {
-#  dtexist <- dtNameStr %in% ls(envir = parent.frame())
-#  if (!dtexist) {
-#    dtDoesNotExistError(dtNameStr, call = call)
-#  }
-#}
+# assertDataTableExists <- function(dtNameStr, call = sys.call(-1)) {
+#   dtexist <- exists(deparse(substitute(dtNameStr)), envir = parent.frame())
+#   #dtexist <- dtNameStr %in% ls(envir = parent.frame())
+#   if (!dtexist) {
+#     dtDoesNotExistError(dtNameStr, call = call)
+#   }
+# }
 
 #' Var Defined?
 #'
--- a/R/asserts.R
+++ b/R/asserts.R
@@ -32,6 +32,22 @@ assertLengthEqual <- function(..., call = sys.call(-1)) {
     }
 }
 
+#' Are arguments equal to value?
+#'
+#' @description Checks if all passed vars are equal to given value.
+#' @param ... Any number of variables as named elements e.g. var1 = var1.
+#' @param val Value to check if variables are equal to
+#' @noRd
+assertEqual <- function(..., val, call = sys.call(-1)) {
+  dots <- dots2argNames(...)
+  
+  notEqual <- !sapply(dots$args, function(i) {i == val})
+  if (any(notEqual)) {
+    notEqualError(dots$names, val, call = call,
+                  msg = "{names *} should be equal to {val}")
+  }
+}
+
 #' Is length correct?
 #'
 #' @description Checks if all passed vars are of length 'length'. Caveat:
--- a/R/asserts.R
+++ b/R/asserts.R
@@ -177,51 +177,6 @@ assertFactor <- function(..., type, call = sys.call(-1)) {
   }
 }
 
-#' Check for Integer OR Factor
-#'
-#' @description Checks if all passed vars and their content are integers or factors.
-#' @param ... Any number of variables as named elements e.g. var1 = var1.
-#' @noRd
-# assertIntegerOrFactor <- function(..., type, call = sys.call(-1)) {
-#   #assertNumeric(..., call = call)
-#   dots <- dots2argNames(...)
-#   #checkIntegerOrFactor <- function(arg) {
-#   #  arg <- unlist(arg)
-#   #  all(assertInteger(var1 = arg) | assertFactor(var1 = arg))
-#   #}
-#   #notInteger <- !sapply(dots$args, checkInteger)
-# 
-# 
-#   #notIntOrFac <- !sapply(dots$args, function(x) {((x == as.integer(x)) | is.factor(x))})
-# 
-#   # checkInteger <- function(arg) {
-#   #   arg <- unlist(arg)
-#   #   all(tryCatch({assertNumeric(arg)},
-#   #                error = FALSE,
-#   #                finally = {
-#   #                  arg == as.integer(arg)
-#   #                }))}
-#   
-#   # checkInteger <- function(arg) {
-#   #   arg <- unlist(arg)
-#   #   all(if (is.numeric(arg)) {
-#   #     arg == as.integer(arg)
-#   #   } else {FALSE})
-#   # }
-#   # 
-#   # 
-#   # notInteger <- !sapply(dots$args, checkInteger)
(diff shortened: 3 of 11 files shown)


R/generate_data.R
```diff
@@ -385,38 +392,68 @@ genFormula <- function(coefs, vars) {
...
-    trimvalue = trimvalue
-  )
+  if (!is.null(startProb)) {
+    dprob <- defData(varname = "prob", formula = startProb, dist = "categorical")
+    dd <- genData(n = n, dprob, id = id)
+    dd <- addMarkov(dd, transMat, chainLen, wide, id,
+      pername, varname, widePrefix,
+      start0lab = "prob",
+      trimvalue = trimvalue
+    )
```
kgoldfeld, line 447: This seems to work quite well, but I when the `startProb` probabilities do not add up to 1, it can generate an error.

```
library(simstudy)

P <- matrix(c(.4, .5, .1,
              .1, .3, .6,
              .3, .4, .3), 
  nrow = 3, byrow = T)

genMarkov(101, P, 2, wide = TRUE, startProb = "0.3;0.3;0.1")
#> Warning: Probabilities do not sum to 1. Adding category to all rows!
#> Error in markovChains(n, transMat, chainLen, s0): Row index is out of bounds: [index=3; row extent=3].
```

`genData` is written so that If you generate a categorical variable with probabilities that add up to less than 1, another category is added (and a warning will be generated). So, in this case, if we specify `"0.3;0.3;0.1", a fourth category with probability .3 will be added:

```
d2 <- defData(varname = "c", formula = "0.3;0.3;0.1", dist = "categorical")
genData(10000, d2)[, table(c)]
#> Warning: Probabilities do not sum to 1. Adding category to all rows!
#> c
#>    1    2    3    4 
#> 3057 2975  986 2982
```

This creates problems for `genMarkov` because there are now 4 start states and the transition matrix only has 3 states - and that is why the error is generated by `genMarkov`. The easiest solution is probably to a specific check in `genMarkov` to make sure the probabilities sum to 1. But, I am open to other solutions.

And this brings to mind another possible error with `addMarkov` - if the user generates start states that don't exist in the transition matrix, errors could be generated. Could you check that out as well?

Just as an aside, if probabilities add up to more than 1, the function normalizes the probabilities to ensure that the probabilities do add up to 1, generates the data, and issues a warning. This does not lead to any problems for `genMarkov`, because there are the correct number of categories.

elinsooon: I believe I addressed both of these things, take a look

kgoldfeld: It definitely looks like you have implemented the check - so that it does indeed issue a warning when the startProb sum is not 1. But when `startProb` is specified, `genMarkov` fails (and your test picks that up). So - this generates an error:

```
genMarkov(101, P, 2, wide = TRUE, startProb = "0.5;0.3;0.2")
```

But this does not:

```
genMarkov(101, P, 2, wide = TRUE)
```

elinsooon: Yep I just caught that and updated it

R/add_data.R
```diff
@@ -299,7 +299,7 @@ addMarkov <- function(dd, transMat, chainLen, wide = FALSE, id = "id",
 
   # if start0lab defined, check that it exists in the transition matrix
   if (!is.null(start0lab)) {
-    if (any(0 < dd[, start0lab, with = FALSE] | dd[, start0lab, with = FALSE] > dim(transMat)[1])) {
+    if (any(0 > dd[, start0lab, with = FALSE] | dd[, start0lab, with = FALSE] > dim(transMat)[1])) {
       c <- condition(c("simstudy::start0probNotInTransMat", "error"),
```
kgoldfeld, line 303: Makes sense why it was failing. The fix seems like it should work, but maybe it should be:

```
any(dd[, start0lab, with = FALSE] < 1 | dd[, start0lab, with = FALSE] > dim(transMat)[1]))
```

i.e. if there is a starting value of 0, it should fail.

elinsooon: Yes sorry I missed that when I was fixing it the first time, updated it

kgoldfeld: I think it is looking good. I will go through your changes one more time, just to confirm everything else is all good. Can you remind me what additional features you've added other than the new `startProb` argument - I want to make sure our `news` file is updated - you could do that if you want.

elinsooon: I just updated genFormula for use with double dot notation, not sure if you've taken a look at that. I think in terms of actually usable additions those are the only two. How do I go about updating the news, I see the file, should I put those things under new features?

kgoldfeld: That would be good.

tests/testthat/test-generate_data.R
```diff
@@ -381,3 +381,156 @@ test_that("genDummy works.", {
...
+  }
+  
+  for (i in 1:5) {
+    check_equal(gm1_nw, gm1)
+    check_equal(gm2_nw, gm2)
+  }
+  
+  set.seed(oldSeed)
+  
+})
```
kgoldfeld, line 536: I like what you did conceptually, but I changed the coding a bit so that it is again less hard coded.

```
set.seed(23456)
gm1_w <- genMarkov(n=nind, transMat = matr, chainLen = nchain, wide = TRUE)

set.seed(23456)
gm2_w <- genMarkov(n=nind, transMat = matr, chainLen = nchain, wide = TRUE, startProb = "0.65;0.25;0.05;0.05")

check_equal <- function (gm_not_wide, gm_wide) {
  rand_id <- sample(n, 1)
  rand_state <- sample(l, 1)
  gmnw <- gm_not_wide[id == rand_id & period == rand_state, state]
  gmw <- gm_wide[rand_id, rand_state + 1, with = FALSE]
  
  expect_equal(as.numeric(gmnw), as.numeric(gmw))
}

for (i in 1:5) {
  check_equal(gm1, gm1_w)
  check_equal(gm2, gm2_w)
}
```

elinsooon: I see I'll put your stuff in. I want to confirm that I understand what you're doing in a few places. You call expect_silent a few times when I think you're looking to check whether an expression is True, for example expect_silent(all(prop[, dif <0.2])). But even if this is False, expect_silent will pass. Is this what you're expecting or am I misunderstanding this.

kgoldfeld: You're right - in that case, it should be `expect_true`.

R/generate_data.R
```diff
@@ -309,30 +310,64 @@ genFormula <- function(coefs, vars) {
...
   }
 
   for (i in 1:(lcoef - 1)) {
-    form <- paste(form, "+", coefs[i], "*", vars[i])
+    #form <- paste(form, "+", coefs[i], "*", vars[i])
+    cg <- coef_gen(coefs[i])
+    form <- paste(form, "+", cg, "*", vars[i])
   }
 
   return(form)
```
kgoldfeld, line 373: This does seem to work well - nice job - but I had another thought (even before looking at the testing). What if we used `genFormula` to be more literal and just generate a formula literally, so that

```
genFormula(c(1, "..x", "..y"), c("a", "b"))
```

just returns `"1 + ..x * a + ..y * b"` and doesn't do any parsing at all. When the formula is applied later on in `defData`, then the the proper checks will be conducted to make sure that the formula is valid. The only thing `genFormula` would care about is whether there are two vectors passed, and that they are either the same length (no intercept) or the first one is one element longer (the intercept). Either one could be any combo of any kind of strings (double dot or not) and numbers. So even something like this would be allowed:

```
genFormula(c(1, "x", "..y"), c("..a", "3"))
```

which would return `"1 + x * ..a + ..y * 3"` or even

```
genFormula(c(1, "..x", "5"), c("..a", "3"))
```

which would return `"1 + ..x *..a + 5 * 3"`

These could all be valid formulas in some data generating process, though the last one is equivalent to  `"16 + ..x *..a"`

I know you spent a lot of time figuring out how to parse the data, but what do you think about this simplification? And maybe it will make the testing easier?

elinsooon: That does make sense to me, it would make the process more efficient for sure between the two functions. This also seems a bit more flexible because you can create a formula with already having the variables defined which could be good in some scenarios. I'll change it to that, getting it to work wasn't too bad, and it was a good exercise, so I really don't mind changing it. And yes, I think it would make the testing easier.

tests/testthat/test-generate_data.R
```diff
@@ -381,3 +381,176 @@ test_that("genDummy works.", {
...
+  # check nFactors are integers
+  expect_error(genMultiFac(1.4, each = 4), class = "simstudy::wrongType")
+  
+  # check length nFactors greater than 2
+  expect_error(genMultiFac(1, each = 4), class = "simstudy::greaterThan")
+  
+  # check number of levels matches factors
+  expect_error(genMultiFac(3, levels = c(2, 3)), class = "simstudy::notEqual")
+  
+  set.seed(oldSeed)
```
elinsooon, line 555: not all of these tests works and I'm not sure why. They pass just find when run in the console and have all the expected outputs, but when I run them as a batch with all the other tests in the file, they give an error and a warning

kgoldfeld: I'll check it out and see if I can uncover anything.

R/conditions.R
```diff
@@ -365,21 +365,64 @@ mismatchError <- function(name_specified, name_null,
...
+
   c <- condition(
-    c("simstudy::dtDoesNotExist", "error"),
+    c("simstudy::notEqual", "error"),
     message, call, ...
   )
   stop(c)
-}
\ No newline at end of file
+}
```
kgoldfeld, line 428: The argument `vars` in `notEqualError was causing me problems, because you want the names, so I've modified the code:

```
notEqualError <- function(names, val, call = sys.call(-1), msg = "", ...) {

   message <- glueCollapse(
         "{ names *} not equal to {val}!", msg
   )

   c <- condition(
      c("simstudy::notEqual", "error"),
       message, call, ...
     )

   stop(c)
}
```

We should really remove the `msg` argument from here, or at least make it work better with the calling function.

NEWS.md
```diff
@@ -1,6 +1,8 @@
 # simstudy (development version)
 
 ## New features
+* Updated genFormula with functionality to generate formulas with double dot variables inside.
```
elinsooon, line 4: I think there was one more thing that I was supposed to put on here but it's slipped my mind

tests/testthat/test-generate_data.R
```diff
@@ -552,5 +552,89 @@ test_that("genMultiFac throws errors.", {
...
+  
+  actual_row_sums1 <- NULL
+  for(i in 1:nrow(g1)) {
+    actual_row_sums1 <- c(actual_row_sums1, sum(g1[i, 2:length(g1)]))
+  }
+  
+  poss_row_sums1 <- sort(unlist(poss_row_sums1))
+  actual_row_sums1 <- sort(unlist(actual_row_sums1))
+  
+  expect_true(all(poss_row_sums1 == actual_row_sums1))
```
elinsooon, line 600: This idea is stemmed from a bit of a logical rabbit hole I went down, I'm not totally sure if it's the best way to test but I wanted to go through with the implementation anyway. Let me know what you think, I can explain my thinking behind why I thought this would work

tests/testthat/test-generate_data.R
```diff
@@ -552,5 +552,89 @@ test_that("genMultiFac throws errors.", {
...
+  
+  
+  # coding == effect, levels == 2
+  nFac <- sample(2:10, size = 1)
+  nEach <- sample(2:10, size = 1)
+  g2 <- genMultiFac(nFac, each = nEach, coding = "effect")
+  for(i in sample(2:length(g2))) {
+    expect_equal(sum(g2[, i, with = FALSE]), 0)
+  }
+  expect_true(!all(g2 == 0))
```
elinsooon, line 610: I've been trying to figure out how to test this case as I have above and below but haven't been able to figure it out. I used Pascal's triangle for the other two but I'm having trouble figuring out the appropriate mathematical relationship for this one.

tests/testthat/test-generate_data.R
```diff
@@ -552,5 +552,89 @@ test_that("genMultiFac throws errors.", {
...
+  
+  actual_row_sums3 <- list()
+  for(i in nrow(g3)) {
+    append(actual_row_sums3, sum(g3[i, 2:length(g3)]))
+  }
+  
+  poss_row_sums3 <- sort(unlist(poss_row_sums3))
+  actual_row_sums3 <- sort(unlist(actual_row_sums3))
+  
+  expect_true(all(poss_row_sums3 == actual_row_sums3))
```
elinsooon, line 633: this test isn't working as it should be right now even though it's passing, I need to iron it out

elinsooon: I've been trying to figure out how to test this case as I have for the first section but haven't been able to figure it out. I used Pascal's triangle for the other one but I'm having trouble figuring out the appropriate mathematical relationship for this one.

tests/testthat/test-generate_data.R
```diff
@@ -612,29 +612,26 @@ test_that("genMultiFac works.", {
...
-  actual_row_sums3 <- list()
-  for(i in nrow(g3)) {
-    append(actual_row_sums3, sum(g3[i, 2:length(g3)]))
+  # checks each column has appropriate number of each value in it (1s, 2s, 3s, etc.)
+  for(i in 2:nFac + 1) {
+    t <- table(g3[, i, with = FALSE])
+    sum_num <- (nLev^(nFac - 1))*nEach
+    
+    expect_true(all(t == sum_num))
+    expect_equal(sum(t), sum_num * nLev)
```
elinsooon, line 633: This would work for the one I use Pascal's Triangle for, let me know if it looks better and if I should change the Pascal's Triangle one too

tests/testthat/test-generate_data.R
```diff
@@ -381,3 +381,257 @@ test_that("genDummy works.", {
...
+  
+  sumTable <- table(rowSums(g3[, 2:length(g3)]))
+  # checks all unique row sums are accounted for
+  for(i in nFac:(nFac * nLev)) {
+    ind <- as.character(i)
+    expect_true(!is.na(sumTable[ind]))
+  }
+  
+  # checks each column has appropriate number of each value in it (1s, 2s, 3s, etc.)
+  for(i in 2:nFac + 1) {
```
kgoldfeld, line 628: This should be `for(i in 2:(nFac + 1)) {`

I still have concerns with this approach, because it seems hard to 'prove' that it would work in all cases. As I said earlier, it seems clear that if the data generation works properly, this will pass. It isn't so obvious that there is no be failed data generation process that could pass. If you can prove - or at least argue very strongly - that _**only**_ a correct randomization will pass - then I will completely trust this. If that proof or argument is not so obvious, then taking another approach may be easier.

One quick thought I had was to convert the sequence of numbers in each case to a single string of numbers, and then use some sort of base-n conversion to convert that sequence to base 10. So, for example, if we have

```
1 1 1
1 1 2
1 1 3
.
.
.
3 3 2
3 3 3
```

Let's say we convert that (by subtracting 1 from everything) to

```
0 0 0
0 0 1
0 0 2
.
.
.
2 2 1
2 2 2
```

Now we have a sequence of numbers in base-3, `000, 001, ... 222`, which can be converted to '0, 1, .. 26`, and if we add 1 to that, we have 1 through 27, and each is unique - we need to generate the data correctly to get that. Maybe this is not a proof, but it is a compelling argument, and I have no doubt that it would pick up errors.

Maybe you could implement something along these lines?

tests/testthat/test-generate_data.R
elinsooon: I updated both of the other scenarios with this method

tests/testthat/test-generate_data.R
elinsooon: I see, I had convinced myself that this was pretty airtight but looking at it more I can see a way where a false generation would pass. What was advantageous about the Pascal's triangle idea is it would just look for the unique sums, but for the number of each sums, which is necessary to prove that the table is generated properly. 

I like the idea you suggested and I'll try writing that in

elinsooon: How does what I did look? I believe it achieves the same idea as what you were saying.

kgoldfeld: I think that could work - even without converting to base 10, as long as you check to make sure there are no categories outside the expected range.

You can replace the loop (which are quite slow in R) with this single `lapply` command. I know it is a little ugly, but it is probably faster:

```
rowStrings <- unlist(lapply(split(g3[,-1], seq(nrow(g3))), function(x) paste0(x, collapse="")))
```

So the whole check reduces to this:

```
nFac <- sample(2:5, size = 1)
nEach <- sample(2:5, size = 1)
nLev <- sample(2:5, size = 1)
g3 <- genMultiFac(nFac, each = nEach, levels = nLev)

rowStrings <- unlist(lapply(split(g3[,-1], seq(nrow(g3))), function(x) paste0(x, collapse="")))
expect_true(length(rowStrings) == (length(unique(rowStrings)) * nEach))
```

Of course, you still need to check that all values are in the correct range ...

elinsooon: I believe I hit both of those

kgoldfeld: I haven't looked at the latest code, but can you clean up, and make sure it includes only the final tests. I guess that means, as cool as the Pascal triangle idea was, you should probably take those out (if you haven't already).

elinsooon: Yes for sure

tests/testthat/test-generate_data.R
```diff
@@ -669,23 +617,25 @@ test_that("genMultiFac works.", {
...
+  
+  # checks all rows are unique
+  rowStrings <- unlist(lapply(split(g4[,-1], seq(nrow(g4))), function(x) paste0(x, collapse="")))
+  expect_true(length(rowStrings) == (length(unique(rowStrings)) * nEach))
+  
+  # checks there is right number of rows
+  expect_true(length(rowStrings) == nEach * prod(nLev))
+  
+  # checks all values are in correct range
+  expect_true(all(g4[, 2:nFac, with = FALSE] <= max(nLev) && g4[, 2:nFac, with = FALSE] > 0))
```
elinsooon, line 638: Everytime I run this test (for all 4 scenarios), it throws an error, not sure what to do about it. It seems like it's an issue with all()? But I could be wrong. The warning for one of them is as follows:
── Warning (test-generate_data.R:618:3): genMultiFac works. ────────────────────
'length(x) = 5625 > 1' in coercion to 'logical(1)'
Backtrace:
 1. testthat::expect_true(...)
      at test-generate_data.R:618:2
 2. testthat::quasi_label(enquo(object), label, arg = "object")
 3. rlang::eval_bare(expr, quo_get_env(quo))

Is the vector being passed to all too long?

kgoldfeld: Weird - I'm not getting any errors. But I do see a couple of flaws in the code. First, you need to look at columns `2 : (nFac + 1)`, or alternatively you need to look at `g4[, -1]`. But, even with that, you really need to evaluate each column separately, because (in this case), there is a maximum value that differs by column.

elinsooon: Maybe it has something to do with my computer processing power, I don't have my desktop to compare it with. I fixed the (nFac + 1) problem in a handful of other places too, I had been making that mistake for a little while. I also addressed your second point.

kgoldfeld: Is it still generating an error on your machine? It wouldn't have anything to do with processing power - it looked like some mismatched vector sizes, or like it was looking for a scalar. But it is weird that it was happening only on your machine. If it is still there, we need to get rid of the error.

elinsooon: Yes I'm still getting it, they're warnings not errors, but that doesn't really make much of a difference

kgoldfeld: That is so strange - I am not getting that - it all seems good. Try to break up the test into two pieces and see where it fails.

elinsooon: Splitting it up seemed to fix it

kgoldfeld (commented):
I think we are getting very close - I mostly have some questions about commented out code and I did find one testing error.

R/asserts.R
```diff
@@ -175,23 +190,34 @@ assertFactor <- function(..., type, call = sys.call(-1)) {
...
 #   
-#   if (any(notInteger && notFactor)) {
+#   notIntOrFac <- !sapply(dots$args, function(x) {all(if(is.numeric(x) {
+#     x == as.integer(x)
+#   }) | is.factor(x)})
+# 
+#   if (any(notIntOrFac)) {
 #     typeError(dots$names[notIntOrFac], type = "integer or factor", call = call)
 #   }
 # }
```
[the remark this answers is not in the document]
elinsooon: This was a niche case assert that I'm not sure we actually need. There's seperate assertInteger and assertFactor functions already. In my opinion I don't think making a combination assertIntegerOrFactor is worthwhile right now, so I'll just delete it.

tests/testthat/test-asserts.R
```diff
@@ -97,16 +104,16 @@ test_that("assertFactor works.", {
...
+# 
 #   expect_silent(assertIntegerOrFactor(
 #     a = as.factor("two"), b = as.factor(123.456),
-#     c = as.factor(c(1, 2, 3)),
-#     d = as.factor(1.1), e = as.factor("one"),
-#     f = 1, g = c(12, 34, 56)
+#     f = 1, g = c(12, 34, 56), 
+#     h = as.factor(c(1, "two", 3))
 #   ))
 # })
```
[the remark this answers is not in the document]
elinsooon: Since I don't think we should build assertIntegerOrFactor, I'll delete these

tests/testthat/test-generate_data.R
```diff
@@ -381,3 +378,282 @@ test_that("genDummy works.", {
...
+  set.seed(98765)
+  
+  # check nFactors are integers
+  expect_error(genMultiFac(1.4, each = 4), class = "simstudy::wrongType")
+  
+  # check length nFactors greater than 2
+  expect_error(genMultiFac(1, each = 4), class = "simstudy::greaterThan")
+  
+  # check number of levels matches factors
+  expect_error(genMultiFac(3, levels = c(2, 3)), class = "simstudy::notEqual")
```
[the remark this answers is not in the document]
elinsooon: Changed to assertLength instead of assertEqual

tests/testthat/test-generate_data.R
```diff
@@ -381,3 +378,282 @@ test_that("genDummy works.", {
...
+  
+  # checks all rows are unique
+  rowStrings <- unlist(lapply(split(g3[,-1], seq(nrow(g3))), function(x) paste0(x, collapse="")))
+  expect_true(length(rowStrings) == (length(unique(rowStrings)) * nEach))
+  
+  # checks there is right number of rows
+  expect_true(length(rowStrings) == nEach * nLev^nFac)
+  
+  # checks all values are in correct range
+  expect_true(all(g3[, i, with = FALSE] <= nLev[i - 1]))
```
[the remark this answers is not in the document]
elinsooon: Since all columns should have the same parameters in terms of maximum and minimum values, I changed this to just observe the whole table (excluding the IDs)

kgoldfeld (commented):
I think it is all good. The only thing that remains is generating the updated md files. Did you change the defaults in Rstudio to oxygenize when you build?

elinsooon:
> I think it is all good. The only thing that remains is generating the updated md files. Did you change the defaults in Rstudio to oxygenize when you build?

It's all set up the way you told me to in an email, are there any extra steps I need to do past that?

kgoldfeld:
So, you need to check to see what version of `Roxygen` you have (probably `7.2.0`) and in the `Description` file, go to the last line and edit the version to the right number so that it is `RoxygenNote: 7.2.0`. And then in the `Build` tab, select `Check`. If your settings are correct, the `.Rd` files should be updated, and you can push those changes to github. Let me know if that doesn't make sense. (You may also get some warnings - if so, let me know what they are.)

elinsooon:
> So, you need to check to see what version of `Roxygen` you have (probably `7.2.0`) and in the `Description` file, go to the last line and edit the version to the right number so that it is `RoxygenNote: 7.2.0`. And then in the `Build` tab, select `Check`. If your settings are correct, the `.Rd` files should be updated, and you can push those changes to github. Let me know if that doesn't make sense. (You may also get some warnings - if so, let me know what they are.)

Got it, I fixed some things that gave warnings that I noted in the recent push, here's the output of everything:

==> Rcpp::compileAttributes()

* Updated R/RcppExports.R

==> devtools::check()

ℹ Updating simstudy documentation
ℹ Loading simstudy
── Building ──────────────────────────────────────────────────────────────────── simstudy ──
Setting env vars:
• CFLAGS    : -Wall -pedantic -fdiagnostics-color=always
• CXXFLAGS  : -Wall -pedantic -fdiagnostics-color=always
• CXX11FLAGS: -Wall -pedantic -fdiagnostics-color=always
• CXX14FLAGS: -Wall -pedantic -fdiagnostics-color=always
• CXX17FLAGS: -Wall -pedantic -fdiagnostics-color=always
• CXX20FLAGS: -Wall -pedantic -fdiagnostics-color=always
────────────────────────────────────────────────────────────────────────────────────────────
✔  checking for file 'C:\Users\User\Desktop\simstudy/DESCRIPTION' ...
─  preparing 'simstudy': (865ms)
✔  checking DESCRIPTION meta-information ... 
─  cleaning src
─  installing the package to build vignettes
✔  creating vignettes (45.3s)
─  cleaning src
─  checking for LF line-endings in source and make files and shell scripts (554ms)
─  checking for empty or unneeded directories
   Removed empty directory 'simstudy/tests/testthat/_snaps'
─  building 'simstudy_0.4.0.9000.tar.gz'
   
── Checking ──────────────────────────────────────────────────────────────────── simstudy ──
Setting env vars:
• _R_CHECK_CRAN_INCOMING_REMOTE_: FALSE
• _R_CHECK_CRAN_INCOMING_       : FALSE
• _R_CHECK_FORCE_SUGGESTS_      : FALSE
• NOT_CRAN                      : true
── R CMD check ─────────────────────────────────────────────────────────────────────────────
─  using log directory 'C:/Users/User/Desktop/simstudy.Rcheck'
─  using R version 4.2.0 (2022-04-22 ucrt)
─  using platform: x86_64-w64-mingw32 (64-bit)
─  using session charset: UTF-8
─  using options '--no-manual --as-cran' (367ms)
✔  checking for file 'simstudy/DESCRIPTION'
─  checking extension type ... Package
─  this is package 'simstudy' version '0.4.0.9000'
─  package encoding: UTF-8
✔  checking package namespace information ...
N  checking package dependencies (7.1s)
   Package suggested but not available for checking: 'formatR'
✔  checking if this is a source package ... 
✔  checking if there is a namespace
✔  checking for executable files (2.5s)
✔  checking for hidden files and directories ... 
✔  checking for portable file names ... 
✔  checking serialization versions ...
✔  checking whether package 'simstudy' can be installed (14.8s)
✔  checking installed package size ... 
✔  checking package directory (528ms)
✔  checking for future file timestamps ... 
✔  checking 'build' directory ...
✔  checking DESCRIPTION meta-information (337ms)
✔  checking top-level files
✔  checking for left-over files ... 
✔  checking index information ... 
✔  checking package subdirectories (443ms)
✔  checking R files for non-ASCII characters ... 
✔  checking R files for syntax errors ... 
✔  checking whether the package can be loaded (513ms)
✔  checking whether the package can be loaded with stated dependencies (414ms)
✔  checking whether the package can be unloaded cleanly (516ms)
✔  checking whether the namespace can be loaded with stated dependencies (412ms)
✔  checking whether the namespace can be unloaded cleanly (513ms)
✔  checking loading without being on the library search path (557ms)
✔  checking dependencies in R code (1.8s)
✔  checking S3 generic/method consistency (1s)
✔  checking replacement functions (514ms)
✔  checking foreign function calls (721ms)
✔  checking R code for possible problems (9.5s)
✔  checking Rd files (717ms)
✔  checking Rd metadata ... 
✔  checking Rd line widths ... 
✔  checking Rd cross-references ... 
✔  checking for missing documentation entries (515ms)
✔  checking for code/documentation mismatches (1.5s)
✔  checking Rd \usage sections (1.5s)
✔  checking Rd contents ... 
✔  checking for unstated dependencies in examples (813ms)
✔  checking line endings in C/C++/Fortran sources/headers
✔  checking pragmas in C/C++ headers and code ... 
✔  checking compilation flags used ...
✔  checking compiled code (17.6s)
✔  checking installed files from 'inst/doc' ... 
✔  checking files in 'vignettes' ... 
✔  checking examples (8s)
✔  checking for unstated dependencies in 'tests' ... 
─  checking tests ...
✔  Running 'testthat.R' [26s] (26.3s)
✔  checking for unstated dependencies in vignettes (340ms)
✔  checking package vignettes in 'inst/doc' ... 
✔  checking re-building of vignette outputs (34.7s)
✔  checking for non-standard things in the check directory ...
✔  checking for detritus in the temp directory
   
   See
     'C:/Users/User/Desktop/simstudy.Rcheck/00check.log'
   for details.
   
   
── R CMD check results ──────────────────────────────────────────── simstudy 0.4.0.9000 ────
Duration: 2m 18.3s

❯ checking package dependencies ... NOTE
  Package suggested but not available for checking: 'formatR'

0 errors ✔ | 0 warnings ✔ | 1 note ✖

R CMD check succeeded

kgoldfeld:
For some reason, it is still not seeing the new `.Rd` files. But, I am going to go ahead and merge. I can fix later if it is still a problem.

PR #159, opened 2022-06-01
genMiss

elinsooon:
Going to keep this branch as a clone of all my previous work, if you merge the old branch and there's still issues, I'll take any code I do here and put it in a fresh simstudy install and close this pull request

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,8 @@
 # simstudy (development version)
 
 ## New features
+* Updated genFormula with functionality to generate formulas with double dot variables inside.
+
 * Added utility functions survGetParams and survParamPlot to aid users in identifying parameters that can be used to generate desired distributions of time to event data.
 
 * Major updates to functions defSurv and genSurv. It is now possible to generate survival outcomes with hazard functions that change over time. In addition, competing risk outcomes can be explicitly generated.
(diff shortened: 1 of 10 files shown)


tests/testthat/test-missing_data.R
```diff
@@ -25,12 +25,68 @@ test_that("defMiss produces correct output", {
...
+  uAvg <- sum(dtAct[, u]) / nrow(dtAct)
+  u <- uAvg
+  col2Exp <- eval(parse(text = as.character(form2)))
+  col2Act <- sum(missMat[, x3]) / nrow(missMat)
+  col2Diff <- abs(col2Act - col2Exp)
+  
+  expect_true(col2Diff < 0.02)
+  
+  ## check empty column
+  expect_true(all(missMat[, u] == 1))
```
elinsooon, line 72: I realize that this is all pretty hard coded but I wanted to get something down so I could get some feedback from you. I feel like the concepts are there but I don't know that this is really the best way to go about testing this.

Also, I'm not understanding how genMiss works when repeated = TRUE very well, and so I'm having trouble coming up with testing ideas.

kgoldfeld: The check for `u` in `missMat` is good. Also could check for `x2` and `x3` - missingness will depend on the actual value of `m` but not `u`. Also maybe want to confirm that `m` is all 0's.

I'll explain the `repeated = TRUE` option.

PR #161, opened 2022-06-03
missing testing

elinsooon:


tests/testthat/test-missing_data.R
```diff
@@ -5,31 +5,139 @@ test_that("defMiss produces correct output", {
...
+  
+  missMatLong <- genMiss(dtTime, defMlong, idvars = c("id","rx"), repeated = TRUE, periodvar = "period")
+  
+  casevec <- NULL
+  for(i in 1:idNum) {
+    tempMM <- missMatLong[id == i]
+    if(tempMM[1, x1] == 1) {
+      casevec <- c(casevec, all(tempMM[, x1] == 1))
+    }
+  }
```
kgoldfeld, line 100: In general, `for` loops in R are super slow, so do just about anything to avoid them (unless you know they will loop a small number of times). This is an alternative, speedier option (anything with data.tables is going to be fast, because data.table is super fast):

```
ids <- missMatLong[period == 0 & x1 == 1, id]
casevec <- missMatLong[id %in% ids, all(x1 == 1)]
```

Also, do we check for the case where `x1` might be 0 at baseline (time 0), implying that all others must be 0 as well?

tests/testthat/test-missing_data.R
```diff
@@ -5,31 +5,139 @@ test_that("defMiss produces correct output", {
...
+    if(tempMM[1, x1] == 1) {
+      casevec <- c(casevec, all(tempMM[, x1] == 1))
+    } else if(tempMM[2, x1] == 1) {
+      casevec <- c(casevec, (all(tempMM[-1, x1] == 1) && tempMM[1, x1] == 0))
+    } else if(tempMM[3, x1] == 1) {
+      casevec <- c(casevec, (all(tempMM[-(1:2), x1] == 1) && tempMM[1:2, x1] == 0))
+    } else if(tempMM[4, x1] == 1) {
+      casevec <- c(casevec, (all(tempMM[-(1:3), x1] == 1) && tempMM[1:3, x1] == 0))
+    }
+  }
```
kgoldfeld, line 124: I think we can improve on this - as you suspected, but I haven't thought about it yet.

tests/testthat/test-missing_data.R
elinsooon: I'll implement it this way.

Ah I didn't realize that case was also true with baseline, I'll put that in.

tests/testthat/test-missing_data.R
```diff
@@ -108,22 +109,76 @@ test_that("genMiss works", {
...
+  n_occur <- data.frame(table(id4.0))
+  id4.0 <- n_occur[n_occur$Freq == 3, ]
+  id4.1 <- missMatLong[period == 3 & y == 1, id]
+  id4 <- intersect(id4.0, id4.1)
+  cv4.0 <- rbind(missMatLong[id %in% id4 & period == 0],
+                 missMatLong[id %in% id4 & period == 1],
+                 missMatLong[id %in% id4 & period == 2])
+  cv4.0 <- cv4.0[, all(y == 0)]
+  cv4.1 <- missMatLong[id %in% id4 & period == 3, all(y == 1)]
+  expect_true(cv4.0 & all(cv4.1 == TRUE))
```
elinsooon, line 175: I think this could probably be cleaned up a bit, but I wanted to show you this idea that does away with the if/else statements and uses data.frames

kgoldfeld:
Will fix some things in another branch.

PR #171, opened 2022-07-27
don't skip_on_cran with covr

### LINKED ISSUE #170: skip_on_cran() conflicting with covr
@assignUser I am trying to understand how and when we should be using  `skip_on_cran()` in the unit testing. I noticed that it is used sometimes, but not always. 

I have found that if `skip_on_cran()` is called, the `covr` report does not pick it up. I know there is a way around this (see [here](https://github.com/r-lib/covr/issues/77), for example), but was wondering why we are using it at all.

assignUser:
Fixes #170

### DIFF
--- a/.github/workflows/covr.yaml
+++ b/.github/workflows/covr.yaml
@@ -21,5 +21,7 @@ jobs:
         with:
           extra-packages: covr
       - name: Test coverage
+        env:
+          NOT_CRAN: true
         run: covr::codecov()
         shell: Rscript {0}


kgoldfeld:
Will this also work when I run `covr::report()` locally?

assignUser:
No but you can do `devtools::test_coverage()` instead and that will set envvar automatically https://devtools.r-lib.org/reference/test.html

kgoldfeld:
Perfect - you always know how to fix anything - too bad you have moved on to bigger and better things - would love to take simstudy to the next level :).

PR #173, opened 2022-09-28
Adding tests 0722

kgoldfeld:


assignUser:
After some debugging I was able to find the problem, the issue with touchstone is caused by {cli} updating, so nothing for you to worry about (for me though xD). I'll look over the PR now.

NEWS.md
```diff
@@ -1,5 +1,9 @@
 # simstudy (development version)
 
+## Minor fixes
+
+* Fixed bug in `trtAssign`
```
assignUser, line 5: Following semver.org this would be a so called 'patch' release bumping the version to 0.5.1  e.g. with `usethis::use_version("patch")`

R/generate_dist.R
```diff
@@ -515,21 +515,11 @@
 .parseUnifFormula <- function(formula, dtSim, n, envir) {
   range <- .splitFormula(formula)
 
-  if (length(range) != 2) {
```
assignUser, line 518: Why the deletion?

R/group_data.R
```diff
@@ -380,6 +380,8 @@ trtAssign <- function(dtName, nTrt = 2, balanced = TRUE,
     dt[, `:=`(.stratum = NULL, .n = NULL)]
 
     if (nTrt == 2) dt[, grpExp := grpExp - 1]
+    dt[, grpExp := as.integer(grpExp)]
```
assignUser, line 383: Looking at the code when `balanced = FALSE` the `grpExp` are generated via C++ and that returns integers anyway?

tests/testthat/test-generate_data.R
```diff
@@ -154,8 +153,7 @@ test_that("ordinal categorical data is generated correctly.", {
 
 
 test_that("non-proportional ordinal categorical data are generated correctly.", {
-  oldSeed <- .Random.seed
-  set.seed(87920)
+  skip_on_cran()
```
assignUser, line 156: You are skipping all tests on cran now? Where there issues last submission?

tests/testthat/test-actual-distributions.R
```diff
@@ -0,0 +1,45 @@
+library(hedgehog)
+freeze_eval <- names(.GlobalEnv)
```
assignUser, line 2: As you are not using hedgehog in this file you can remove lines 1,2 and 45 :)

tests/testthat/test-actual-distributions.R
```diff
@@ -0,0 +1,45 @@
...
+  def <- defData(varname = "b1", formula = "..p1", variance = 1, dist = "beta")
+  def <- defData(def, varname = "b2", formula = "..logitp2", variance = 1, 
+                 dist = "beta", link="logit")
+  
+  dd <- genData(n = 1000, dtDefs = def)
+  diff <- dd[, .(abs(mean(b1) - p1), abs(mean(b2) - p2)) ]
+  expect_true(all(diff < 0.05))
+  
+  diffv <- dd[, .(abs(var(b1) - v1), abs(var(b2) - v2)) ]
+  expect_true(all(diffv < 0.02))
```
assignUser, line 23: I have not looked through this thoroughly, would need some time but trust you have this aspect covered :+1:

tests/testthat/test-generate_data.R
```diff
@@ -486,15 +481,14 @@ test_that("genMarkov works.", {
...
-  set.seed(23456)
   gm2 <- genMarkov(n = nind, transMat = matr, chainLen = nchain, startProb = "0.65;0.25;0.05;0.05")
 
   prop <- gm1[, .N, keyby = .(id, state)]
   prop[, p := N / sum(N), keyby = id]
   prop[, t.p := rep(theoretical_p, max(id))]
   prop[, dif := abs(p - t.p)]
 
-  expect_true(all(prop[, dif < 0.02]))
+  expect_true(all(prop[, dif < 0.03]))
```
assignUser, line 491: Is the larger dif enough to work without a seed?

tests/testthat/test-generate_data.R
```diff
@@ -705,3 +693,113 @@ test_that("genSynthetic works.", {
...
+                          knots = knots, degree = 3,
+                          noise.var = .025), regexp = "newvar must be a string")
+  expect_error(genSpline( dt = dt, newvar = "weight",
+                          predictor = "x1", theta = theta1,
+                          knots = knots, degree = 3, newrange = "4;3;2",
+                          noise.var = .025), regexp = "Range not specified as two values")
+  # expect_error(genSpline( dt = dt, newvar = "weight",
+  #                         predictor = "x1", theta = theta1,
+  #                         knots = knots, degree = 3, newrange = "1;a",
+  #                         noise.var = .025), regexp = "Non-numbers entered in range")
```
assignUser, line 803: If this test does not work I would remove it xor create a github issue to fix the underlying problem

tests/testthat/test-generate_dist.R
```diff
@@ -70,14 +70,19 @@ test_that("'uniform' formula checked correctly", {
...
+      expect_silent(defData(varname = "z", formula = x$range, dist = "uniform"))
     }
   )
 
-  expect_warning(.genunif(3, "1;1", NULL, environment()), "are equal")
-  expect_error(.genunif(3, "", NULL, environment()), "format")
-  expect_error(.genunif(3, "1;2;3", NULL, environment()), "format")
-  expect_error(.genunif(3, "2;1", NULL, environment()), "'max' < 'min'")
+  expect_error(defData(varname = "z", formula = NULL, dist = "uniform", "format"))
+  expect_error(defData(varname = "z", formula = "1;2;3", dist = "uniform"), "format")
```
assignUser, line 78: The idea here was to test the different .gen* functions directly but your approach is actually correct as testing internal functions is discourage to prevent testing implementations instead of apis/behaviours. so :+1:

assignUser (approved):
I haven't looked at the test logic in detail, I trust you have that well in hand. Otherwise just a few comments.

Great to see how you embrace testing :partying_face: Let me know if there is anything else I might do to help this release along.

R/generate_dist.R
kgoldfeld: That is so strange, I did not intend to do that. I will add back.

R/group_data.R
kgoldfeld: Yes - `trtObserve` is fine for the reason you stated.

tests/testthat/test-actual-distributions.R
kgoldfeld: Done.

tests/testthat/test-actual-distributions.R
kgoldfeld: Not sure I would trust me, but I did vet this pretty thoroughly.

tests/testthat/test-generate_data.R
kgoldfeld: It was a while ago - I don't remember exactly why I started doing this. I think there was an issue, and this was a simple work around. Do you think it is bad practice?

tests/testthat/test-generate_data.R
kgoldfeld: Yes - I think I ran something like 1 million iterations and felt comfortable with that range. There's probably a better way to test this, but that was what I came up with.

tests/testthat/test-generate_data.R
assignUser: Well it kind of subverts the testing process on CRAN... but you are certainly not the only one doing it so :shrug:

tests/testthat/test-generate_data.R
kgoldfeld: The test works but it also generates a warning. I will keep commented out, but will add to issues.

tests/testthat/test-generate_data.R
kgoldfeld: I'll keep them in for now, and can revisit later.

tests/testthat/test-generate_dist.R
kgoldfeld: Yeah - I realize that this was a fundamental issue, and I wasn't certain how to handle it. Once I started doing this, I noticed that this might require many more changes. And it was at this point, that I just stopped altogether, because I wasn't really sure about the way to proceed.

kgoldfeld:
Should I go ahead and merge, or should I wait until you fix the Continuous Benchmarks and make sure it passes that?

assignUser:
I tested the benchmark locally and that looked fine :+1:  it will take me a bit to fix it upstream, so go ahead and merge :)

PR #177, opened 2022-11-07
Correlated data gen

kgoldfeld:


kgoldfeld:
@assignUser Hope you are well. I believe the failure on the continuous benchmarks is a known problem. Is that something I can fix?

assignUser:
I am well thank you! 
Yes this is still the same issue: https://github.com/kgoldfeld/simstudy/pull/173#issuecomment-1261373328 nothing you can do, this is on me ^^

R/generate_correlated_data.R
```diff
@@ -586,3 +586,115 @@ genCorOrdCat <- function(dtName, idname = "id", adjVar = NULL, baseprobs,
...
+  names(blocks) <- paste0("M", c(1 : nPeriods))
+  
+  ### put blocks together
+  
+  block_str <- sapply(nPeriods:1, 
+                      function(x) data.table::shift(1:nPeriods, nPeriods - x, fill = 0))
+  block_str[upper.tri(block_str)] = t(block_str)[upper.tri(block_str)]
+  block_str <- matrix(paste0("M", block_str), nPeriods)
+  
+  bm <- blockmatrix::blockmatrix(value = block_str, list = blocks, dim=c(nPeriods, nPeriods))
```
assignUser, line 697: As {blockmatrix} is only a suggests you should guard this function with `requireNamespace` see: https://r-pkgs.org/dependencies.html#guarding-the-use-of-a-suggested-package

R/generate_correlated_data.R
```diff
@@ -586,3 +586,115 @@ genCorOrdCat <- function(dtName, idname = "id", adjVar = NULL, baseprobs,
...
+  } else if (is.null(iRho)) {
+    irvec <- c(1, rvec[2:nPeriods])
+  }
+  
+  ### generate blocks
+  
+  blocks <- lapply(1:nPeriods, function(x) matrix(rvec[x], nInds, nInds))
+  blocks <- lapply(1:nPeriods, function(x) assignDiag(blocks[[x]], irvec[x]))
+  
+  diag(blocks[[1]]) <- 1
```
assignUser, line 687: Should this not use:
```suggestion
  blocks <- assignDiag(blocks, 1)
```

R/generate_correlated_data.R
```diff
@@ -586,3 +586,115 @@ genCorOrdCat <- function(dtName, idname = "id", adjVar = NULL, baseprobs,
...
+  }
+  
+  assertOption(corstr = corstr, options = c("ind", "cs", "ar1"))
+  
+  ###
+  
+  assignDiag <- function(block, value) {
+    diag(block) <- value
+    return(block)
+  }
```
assignUser, line 650: Not sure if this was your intention but I like how this function turns the "impure" (because it modifies the matrix in-place) `diag` function into a pure function :+1:

assignUser (approved):
A few minor suggestions

R/generate_correlated_data.R
kgoldfeld: Done - thanks.

R/generate_correlated_data.R
kgoldfeld: Actually, I had restructured the code so that I didn't need that extra step. I meant to remove it completely - and it is gone now.

R/generate_correlated_data.R
kgoldfeld: I definitely had not been aware of the formal distinction between "pure" vs "impure" functions, but I have been wary of the kinds of functions that are impure, because it does make it feel like things are a bit out of control. The package "data.table" functions seem to have long lasting effects on data sets when used in a function. I don't quite see how what I've done here makes the overall genBlockMat pure, since I could have put assignDiag outside the function and I would not have been modifying anything outside the function addCorGen. I did it, because it seems to be a pretty specific case where I needed to use lapply with diag, and this was the only way I could get it to work properly. But, I'll take credit for being "pure". :).

PR #178, opened 2022-11-16
Fix add cor gen 2

kgoldfeld:
This update involves some major changes to `addCorGen` and moves some R code to Rcpp to improve performance substantially.

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -124,25 +124,6 @@ references:
   year: '2022'
   url: https://CRAN.R-project.org/package=mvnfast
 - type: software
-  title: mvtnorm
-  abstract: 'mvtnorm: Multivariate Normal and t Distributions'
-  notes: Imports
-  authors:
-  - family-names: Genz
-    given-names: Alan
-  - family-names: Bretz
-    given-names: Frank
-  - family-names: Miwa
-    given-names: Tetsuhisa
-  - family-names: Mi
-    given-names: Xuefei
-  - family-names: Hothorn
-    given-names: Torsten
-    email: Torsten.Hothorn@R-project.org
-    orcid: https://orcid.org/0000-0001-8301-0471
-  year: '2022'
-  url: https://CRAN.R-project.org/package=mvtnorm
-- type: software
   title: Rcpp
   abstract: 'Rcpp: Seamless R and C++ Integration'
   notes: Imports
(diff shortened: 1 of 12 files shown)


kgoldfeld:
@assignUser I made some pretty big changes to `addCorGen` that speed things up (moved some functions to Rcpp) and make it more flexible. I am quite sure I didn't break anything in the process. There weren't a lot of tests in this part of the code, and I didn't get to it here. I'll do that soon, but need to get this out sooner.

There was one problem with a number of servers (windows oldrel and macos oldrel) that couldn't find the package `pbv` which I used in the new Rcpp function for the `ep` method. For some reason that I don't understand, this was a problem with one of my examples. I removed that example, and everything is good. But that didn't really solve the problem - I'll note is an issue and try to fix at some point before we move to CRAN. 

If this all sounds good, I'll merge for now.

assignUser:
@kgoldfeld I won't have time before the weekend to look this over in detail. Feel free to merge if this is blocking you or users. I can still add comments on a closed PR :) I will also fix the benchmarking action this weekend so we can see the improvements! :rocket:

kgoldfeld:
That sounds good. We can always fix/clean any more issues later. Thanks.

assignUser:
@kgoldfeld I have fixed https://github.com/lorenzwalthert/touchstone/issues/118 so the benchmark workflow should be fixed now :+1:

kgoldfeld:
That is great to hear. I was wondering how to deal with the issue that at least two of the servers that we are running R CMD CHECK on are giving an error indicating they are not finding the `pbv` package (or last a specific function within that package). Is it possible that the server doesn't have the latest version of that package loaded? If not, do you have any idea how we could make that happen? I am afraid if I develop any tests for the updated `genCorGen` and `addCorGen` functions, they will generate errors in the R CMD CHECK. Do you know who maintains those servers, or are they not even actual servers, but something virtual?

assignUser:
I have check out the logs,the latest CRAN version of  `pbv` is installed but for some reason on oldrel and devel a specific Rcpp function of the package can not be found... I have little experience with Rcpp so I am not sure what the issue could be but there is a more recent version of the package on github. I look into it on the weekend.

assignUser:
@kgoldfeld this seems to happen only with oldrel (so R 4.1.3) you could test if that issue also happens locally with that r version and `devtools::check()` (or test() should even be enough?)

kgoldfeld:
Thanks - I'll give it a try.

PR #181, opened 2023-01-23
Fixing genSpline error checking

kgoldfeld:
Made some changes to fix issue.

### DIFF
--- a/R/generate_data.R
+++ b/R/generate_data.R
@@ -888,33 +888,27 @@ genSpline <- function(dt, newvar, predictor, theta,
 
   # Check arguments
 
-  if (!exists(deparse(substitute(dt)), envir = parent.frame())) {
-    stop("Data table does not exist.")
-  }
-
-  if (!predictor %in% names(dt)) {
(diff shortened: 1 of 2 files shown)


kgoldfeld:
I am merging even though there is a warning from the mac os servers related to a warning issued for `sprintf`:

`/Users/runner/work/_temp/Library/Rcpp/include/Rcpp/internal/r_coerce.h:255:7: warning: 'sprintf' is deprecated: This function is provided for compatibility reasons only.  Due to security concerns inherent in the design of sprintf(3), it is highly recommended that you use snprintf(3) instead. [-Wdeprecated-declarations]`

This is unrelated to this merge - will need to resolve in the future in any case.

assignUser:
@kgoldfeld this is an issue already fixed by Rcpp and is in the latest cran release which actually happened today. So on the next PR this issue should be solved already :tada:

kgoldfeld:
Yes that is what I figured, as I am getting no issues on my mac with the latest version of Rcpp. And I can see that snprintf is used.

PR #182, opened 2023-01-31
Issue 180 flexible corr matrix

kgoldfeld:
Updated genCorMat and added tests


kgoldfeld:
@assignUser I've made some enhancements to `genCorMat`. Everything seems to be working fine, and I was going to merge. Not sure if you are interested in taking a look before I do that. No worries if you don't have the time.

assignUser (commented):
@kgoldfeld I gave it a glance for now and nothing major jumped at me. I really like how you have embraced the error system and testing!

kgoldfeld:
@assignUser Thanks for taking a look. Yes, I've tried to expand the error checks as needed so that things remain sort of consistent. I do make sure the testing is completed for anything new I add, though I doubt I am creating the most robust testing suite - I can see that is a whole art/science unto itself. 

I am happy to see that all the checks are now passing on all the servers - glad to see things have been updated on the server end.

kgoldfeld:
@assignUser I actually have another question for you. In the current development version, there is a new function `genBlockMat`. That has been there for a few months. I wasn't actually totally happy with it, and I went ahead and changed it pretty substantially - the function arguments are quite different. I was wondering how uncool that is to break something that is in development, or is that just the risk someone takes when they use it. (I did write about it on my blog, so there is a chance someone is using it.)

kgoldfeld:
@assignUser Related to the question about breaking `genBlockMat` I have another. There are actually two different structures for these block matrices that have slightly different parameterization, and slightly different arguments. I am wondering if it actually makes sense to split them up into separate functions.

The first structure is considered "exchangeable" (I would call in `matBlockExchange`, and uses $\rho_w$ (for within period correlation), $\rho_b$ (for between period correlation), and $\rho_a$ (for within individual correlation or auto-correlation). The second structure is considered a "decay" structure (`matBlockDecay`), and uses only  $\rho_w$ and $r$ (the decay parameter). Both structures have a *cross-sectional* vs *cohort* option, which influences how those parameters are applied. (For example $\rho_a$ is only used in the *cohort exchangeable* structure, and $r$ is used differently in the *cross-sectional decay* model compared to the *cohort decay* model.)

As I mentioned, I have combined them in the new version of `genBlockMat`, but in using it a bit writing up a vignette, I am wondering how confusing it might be in its combined form. I was wondering what you think of the tradeoff between a single function with more complexity compared with the slightly more complicated set of two but slightly simpler functions. (And in terms of coding, they would actually share a common core functionality that I could pull out of both, so that they can share.)

Do you have an opinion, or does it not really matter much?

kgoldfeld:
Thanks - I appreciate the feedback - so I will continue to develop the two somewhat simpler functions. As for keeping a package simple, I think that ship has sailed. No way to fix that, unless we start over #simstudy2.

assignUser:
> As for keeping a package simple, I think that ship has sailed. No way to fix that, unless we start over #simstudy2.

Well it's not rare for these principles to be aspirational :D

kgoldfeld:
@assignUser The CITATION.cff update is failing. Is that something we can deal with, or is that a server issue that will just resolve itself with time?

assignUser:
@kgoldfeld that workflow uses old version of the actions. I will put up a PR to udpate them.

Issue #192, opened 2023-03-30
add double-dot functionality for defSurv


kgoldfeld:
A user has requested that we add double-dot functionality (dynamic definitions) to `defSurv` - that sounds like a good idea.

PR #193, opened 2023-04-03
Adding double-dot functionality to defSurv

kgoldfeld:
Addressing Issue #192

### DIFF
--- a/R/generate_data.R
+++ b/R/generate_data.R
@@ -1035,27 +1035,38 @@ genSurv <- function(dtName, survDefs, digits = 3,
   
   for (i in (seq_along(events))) {
     
-    nlogu <--log(stats::runif(nrow(dtSurv), min = 0, max = 1))
+    nlogu <- -log(stats::runif(nrow(dtSurv), min = 0, max = 1))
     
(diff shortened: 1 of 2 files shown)


R/generate_data.R
```diff
@@ -1035,27 +1035,38 @@ genSurv <- function(dtName, survDefs, digits = 3,
   
   for (i in (seq_along(events))) {
     
-    nlogu <--log(stats::runif(nrow(dtSurv), min = 0, max = 1))
+    nlogu <- -log(stats::runif(nrow(dtSurv), min = 0, max = 1))
```
assignUser, line 1038: An important space :D

assignUser (commented):
Nice!  :rocket:

R/generate_data.R
kgoldfeld: @assignUser Very important. Interestingly, data.table has literally implemented its own double-dot notation that does virtually the same thing ours does. There's actually no need to parse it, data.table does it for you. But, I just used ours for consistency sake.

assignUser: Oh neat I did not know that, dt was the inspiration for the `..` prefix  :D

Issue #194, opened 2023-04-03
Generate unbalanced cluster sizes


kgoldfeld:
It could be helpful to relax the assumption of exactly balanced cluster size when estimating statistical power using simulation, so I propose that we add a new "distribution" called something like _clusterSize_ that will implement this. As part of data definitions, the user will specify the overall sample size in the _formula_ argument, and will input a _dispersion_ parameter (that will be input into the _variance_ field that is a non-negative number, with 0 meaning exact balance, and larger values implying more variability in the cluster sizes. The only _link_ option argument will be "identity" (the default, so does not need to be specified).

kgoldfeld:
Done.

PR #195, opened 2023-04-03
Adding clusterSize distribution

kgoldfeld:
This is addressing Issue #194

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -7,8 +7,10 @@ double-dot variables in scale, shape, and formula parameters.
 distribution in `defData` and `defDataAdd`.
 
 ## Major fix
-* Updated the random effect variance generation for function `iccRE` under the
-Poisson distribution.
+* Improved the random effect variance generation for function `iccRE` under the
+Poisson distribution. The current approach is based on the 2013 paper by
+Nakagawa & Schielzeth titled "A general and simple method for obtaining $R^2$ from
+generalized linear mixed-effects models"
 
 ## Minor fix
 * Modified internal function to speed up beta distribution data generation.
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,8 @@
 ## New features
 * Added double-dot (dynamic) functionality to defSurv. Users can now specificy 
 double-dot variables in scale, shape, and formula parameters.
+* It is possible to generate variable cluster sizes using the *clusterSize*
+distribution in `defData` and `defDataAdd`.
 
 ## Minor fix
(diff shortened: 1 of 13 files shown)


kgoldfeld:
@assignUser I don't know if you want to take a look at this before I go ahead and merge, but I added a simple distribution called _clusterSize_. I'm thinking it will actually be useful (at least for me).

NEWS.md
```diff
@@ -1,8 +1,14 @@
...
 ## New features
-* Added double-dot (dynamic) functionality to defSurv. Users can now specificy 
+* Added double-dot (dynamic) functionality to defSurv. Users can now specify 
 double-dot variables in scale, shape, and formula parameters.
+* It is possible to generate variable cluster sizes using the *clusterSize*
+distribution in `defData` and `defDataAdd`.
+
+## Major fix
+* Updated the random effect variance generation for function `iccRE` under the
+Poisson distribution.
```
assignUser, line 11: What was the issue here? Was there a bug that might have influenced existing code/results?

R/generate_dist.R
```diff
@@ -595,3 +601,27 @@
...
+    if (d == 0) {
+      ss <- floor(total/n)
+      s <- rep(ss, n)
+      surplus <- total - ss*n
+      if (surplus > 0) {
+        upgrade <- sample(1:n, surplus, replace = FALSE)
+        s[upgrade] <- s[upgrade] + 1
+      }
+    } else {
+      x <- dirmult::rdirichlet(1, alpha = rep(1/d, n) )[1,]
```
assignUser, line 622: You added `dirmult` as a Suggests so you should guard this function to error gracefully if the package is not installed (I think we have done this in other functions too?)

NEWS.md
assignUser: Looking at the code below "improved" might be clearer  than updated?

assignUser (approved):
Some minor comments :)

NEWS.md
kgoldfeld: Yeah - I was trying to find the source of my code - it looked really odd (though I know I did it). I found the source that I thought I was using (it is now referenced in the iccRE.md file), and it was quite different and simpler than what I was doing. It also has the benefit of being correct. Though the ICC is not such an obvious concept in the context of the Poisson distribution, so I am not sure if the one I generated was wrong or just a different definition. This isn't great, but I just can't figure out where the old method came from.

kgoldfeld:
@assignUser Did you see the other `clusterSize` addition, and it all looked reasonable?

PR #214, opened 2023-08-29
Fix values generated by trtAssign

### LINKED ISSUE #213: Treatment values change when `ratio` argument is used?
I am confused by the following result. Why does adding the `ratio` argument change the treatment label from 0,1 to 1,2?

Thank you for creating this package, just curious to know if I'm missing something here!

```r
library(simstudy)
library(data.table)
```

### Input
```r
dd <- genData(10)
dd <- trtAssign(dd,
                nTrt = 2,
                ratio = c(1,3),
                balanced = FALSE,
                grpName = "tx")

print(dd$tx)
```

### Output
```
 [1] 1 2 2 1 1 2 1 2 2 1
```

### Input
```r
dd <- genData(10)
dd <- trtAssign(dd,
               nTrt = 2,
               #ratio = c(1,3),
               balanced = FALSE,
               grpName = "tx")

print(dd$tx)
```

### Output
```
[1] 0 1 0 1 0 0 0 1 0 0
```

assignUser:
Closes #213

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # simstudy (development version)
 
+## Breaking Changes
+* `trtAssign` with `ratio=Null` used to produce 0-index values but 1-indexed values if `ratio` was set. This was adjusted so now both versions produce 1-indexed values. This is a potentially breaking change for existsing scripts that use the generated treatment values while assuming the old behavior (e.g. using hardcoded values to filter).
+
 ## New features
 * Function `logisticCoefs` determines the intercept and treatment/exposure 
 parameter for a data generating process (based on a logistic regression model) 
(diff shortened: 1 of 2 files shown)


kgoldfeld:
@assignUser The proposed fix changes the case where `balance = FALSE` so that 2 treatment categories to generates 1/2, as opposed to 0/1. I think we want to make sure the result is always 0/1 with two treatments.

assignUser:
ah yeah also makes sense, that only requires a small change to `.stratSample` I think :)

kgoldfeld:
I appreciate your jumping in to try to fix this, but if you'd like me to take it from here, I am happy to.

assignUser:
After another look `.stratSample` is of course not involved as `balanced=FALSE` so I feared the issue is in the cpp code but actually it'shttps://github.com/kgoldfeld/simstudy/blob/d40d33d4c869ff568608d828a11ea60c0b20a6d9/R/group_data.R#L534 
So should we just -1 the whole col regardless of `length(formulas)?

PR #215, opened 2023-08-30
Fix trt assign

kgoldfeld:
@assignUser I created this new pull request. Not sure this is proper workflow, but it is what it is. Check it out. It seems to do the trick.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # simstudy (development version)
 
+## Breaking Changes
+* `trtAssign` with `ratio=Null` used to produce 0-index values but 1-indexed values if `ratio` was set. This was adjusted so now both versions produce 1-indexed values. This is a potentially breaking change for existsing scripts that use the generated treatment values while assuming the old behavior (e.g. using hardcoded values to filter).
+
 ## New features
 * Function `logisticCoefs` determines the intercept and treatment/exposure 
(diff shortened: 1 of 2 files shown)


NEWS.md
```diff
@@ -1,5 +1,8 @@
 # simstudy (development version)
 
+## Breaking Changes
+* `trtAssign` with `ratio=Null` used to produce 0-index values but 1-indexed values if `ratio` was set. This was adjusted so now both versions produce 1-indexed values. This is a potentially breaking change for existsing scripts that use the generated treatment values while assuming the old behavior (e.g. using hardcoded values to filter).
```
assignUser, line 4:
```suggestion
* `trtAssign` with `ratio=NULL` used to produce 0-index values but 1-indexed values if `ratio` was set. This was adjusted so now both versions produce 0-indexed values. This is a potentially breaking change for existing scripts that use the generated treatment values while assuming the old behavior (e.g. using hardcoded values to filter).
```

assignUser (approved):
Yep totally fine to open another PR for the same thing! Looks fine to me. Though this leaves `trtObserve` to produce 0/1 for singular formulas and 1...n for n formulas but I guess that makes sense withing the context of that function and is also mentioned in the docs:
> @return An integer (group) ranging from 1 to length of the probability vector

PR #217, opened 2023-11-22
Updating missing test

kgoldfeld:


kgoldfeld:
This is a minor fix to a test that failed one of the servers on CRAN in a freak event. I just increased the sample size of the test to reduce the likelihood of such an error in the future.

PR #223, opened 2024-04-12
222 create a longitudinal dataset is not just counting integers but reflects actual time points

kgoldfeld:


kgoldfeld:
@assignUser I made a relatively simple addition, and fixed ups some things in the `addPeriods` function, including improving the help documentation. If you want to take a look before I merge, I will hold off.

assignUser (approved):
Very nice, I can see how this opens new use cases!

PR #225, opened 2024-05-10
216 nonrandom distribution returns a single value when repeated values are expected

kgoldfeld:


kgoldfeld:
I figured out the test issue - needed to change the environment when in the `do.call()`. It has the added benefit of working if the custom function is defined within another function. Let me know if you think I can merge this.

R/generate_dist.R
```diff
@@ -336,6 +345,34 @@
...
+# @param args String of comma delimited arguments that will be passed to fn
+# @param dtSim Incomplete simulated data.table
+# @return A data.frame column  with the updated simulated data
+
+.gencustom <- function(n, fn, args, dtSim, envir) {
+  
+  # new <- .evalWith(formula, .parseDotVars(formula, envir), dtSim, n)
+  
+  args <- gsub("\\s+", "", args) # remove any spaces
+  arg_l <- as.list(unlist(strsplit(args, ",")))
```
assignUser, line 361:
```suggestion
   arg_l <- args |> strsplit(",") |> unlist(recursive = FALSE) |> trimws()
```
with recursive it only strips one layer of list, that makes `as.list` superfluos, `trimws` removes leading and trailing whitespace. That means it wont remove the spaces around = if that's passed in, not sure if that works with the rest of the code, so maybe `gsub` is better :)

R/generate_dist.R
```diff
@@ -336,6 +345,34 @@
...
+  arg_l <- lapply(arg_l, function(a) as.list(unlist(strsplit(a, "="))))
```
assignUser, line 362: Ah, just one line below my question is answered :D so it either needs trimws here too or the gsub above.

assignUser (approved):
Wow, very cool, nice job! Neat how the systems come together and something is easier than expected 😁 

I have just two small question:
- this will at the moment only work with arguments fully specified as `name = vale||var` (just from looking at the code) If I am not mistaken I think that should be mentioned in the documentation to avoid confusion. (or the code changed to accept a mix as well but I think it's totally fine keeping it as is)
- double-dot variables should also work right? that might also be worth a mention.

R/generate_dist.R
kgoldfeld: I really need to strip out all spaces - and create a list of lists of two items (the variable name and the formula) - so I think I'll keep it as I wrote it.

kgoldfeld:
> Wow, very cool, nice job! Neat how the systems come together and something is easier than expected 😁
> 
> I have just two small question:
> 
>     * this will at the moment only work with arguments fully specified as `name = vale||var` (just from looking at the code) If I am not mistaken I think that should be mentioned in the documentation to avoid confusion. (or the code changed to accept a mix as well but I think it's totally fine keeping it as is)

Yes - I will make it clear that we need that format "name = value or formula"

> 
>     * double-dot variables should also work right? that might also be worth a mention.

Yes - double dot works here as well - that was the beauty of the system you set up - I was able to take advantage here.

kgoldfeld:
I have updated the help documentation in [simstudy.Rmd](https://github.com/kgoldfeld/simstudy/pull/225/commits/eb4175b503eaca551f575198b7032d829bef80cc).

One thought I had. I now have the function name in the defData *formula* field, and the formula string in the *variance* field, becuase I thought it was logical to have the function name first. However, it might be less confusing to switch them so that the formula string is in the defData *formula* field, and the function name in the defData *variance* field. Do you think the change would be useful?

assignUser:
>Do you think the change would be useful?

I think either way works but I would also tend to put the function name first, it just feels 'natural' even if the field name doesn't fit ^^

kgoldfeld:
I guess I landed in the same place as you. This one is in the books - since it has been merged. I will likely be submitting the current development version as simstudy 0.8.0 in the next few weeks.

Issue #210, opened 2023-07-28
genCorGen with varying cluster sizes


kgoldfeld:
`addCorGen` cannot handle variable cluster sizes if a correlation and correlation structure are specified. This can currently be addressed by using `genCorMat` to generate a list of variable size matrices. However, this could be automated within `addCorGen`, so that the variable size matrices are generated.

PR #229, opened 2024-06-05
210 gencorgen with varying cluster sizes

kgoldfeld:
@assignUser I addressed #210 by updating function `addCorGen`. As part of this update, I fleshed out the help details. Here they are so you can understand what I did:

> The original data table can come in one of two formats: a single row per **idvar** (where data are *ungrouped*) or multiple rows per **idvar** (in which case the data are *grouped* or clustered). The structure of the arguments depends on the format of the data.
>  
> In the case of *ungrouped* data, there are two ways to specify the number of correlated variables and the covariance matrix. In approach (1), **nvars** needs to be specified along with **rho** and **corstr**. In approach (2), **corMatrix** may be specified by identifying a single square *n* x *n* covariance matrix. The number of new variables generated for each record will be *n*. If **nvars**, **rho**,  **corstr**, and **corMatrix** are all specified, the data will be generated based on the information provided in the covariance matrix alone. In both (1) and (2), the data will be returned in a wide format.
>  
> In the case of *grouped* data, where there are *G* groups, there are also two ways to proceed. In both cases, the number of new variables to be generated may vary by group, and will be determined by the number of records in each group, $n_i, i \in \{1,...,G\}$ (i.e., the number of records that share the same value of *idvar*). **nvars** is not used in grouped data. In approach (1), the arguments **rho** and **corstr** may both be specified to determine the structure of the covariance matrix. In approach (2), the argument **corMatrix** may be specified. **corMatrix** can be a single matrix with dimensions $n \ \text{x} \ n$ if $n_i = n$ for all *i*. However, if the sample sizes of each group vary (i.e., $n_i \ne n_j$ for some groups *i* and *j*), **corMatrix** must be a list of covariance matrices with a length *G*; each covariance matrix in the list will have dimensions $n_i \ \text{x} \ n_i, \ i \in \{1,...,G\}$. In the case of *grouped* data, the new data will be returned in *long* format (i.e., one new column only).

I didn't add any tests here. I need to do this, because this part of the package is woefully untested; I will hopefully get to it when I have more time.


NEWS.md
```diff
@@ -1,9 +1,15 @@
 # simstudy (development version)
 
+## New features
```
assignUser, line 3: This confused me at first because I was looking for the associated change xD

R/add_correlated_data.R
```diff
@@ -527,7 +597,7 @@ addCorGen <- function(dtOld, nvars=NULL, idvar = "id", rho=NULL, corstr=NULL, co
     }
     
     dX <- dtTemp[, list(.id, seq_, .XX)]
-
+
```
assignUser, line 600:
```suggestion

```
some superfluous whitespace

R/add_correlated_data.R
```diff
@@ -457,24 +511,41 @@ addCorGen <- function(dtOld, nvars=NULL, idvar = "id", rho=NULL, corstr=NULL, co
...
-
-  ####
+  # nvars <- dtTemp[.id == 1, .N] # only permits case where number of records per id is the same
+  
+  counts <- dtTemp[, .N, by = .id][, N]
+  same_nvar <- all(counts == counts[1])
+  
+  if (!wide) {    # multiple record per id
+    if (is.null(corMatrix)) {
+      if (same_nvar) {
```
assignUser, line 538: It's fine here imo because it's just the one line each but when you nest three levels deep it's usually a good time to check if pulling some of the logic out can flatten things out a bit. (a goto example: instead of wrapping the entire function with `if(condition)` invert the condition and return early removing that level of nesting from the entire function)

assignUser (approved):
:+1: 
>  this part of the package is woefully untested

Yeah it's surprisingly hard to retroactively add tests to things ^^

PR #233, opened 2024-07-19
232 functions used in formula= argument in defdata not available for functions defined out of global namespace

kgoldfeld:


kgoldfeld:
@assignUser Everything seems to be good, except there is this very strange issue with one of the vignettes (double_dot_extension.Rmd) that fails to build during `R CMD check`.  The message is kind of odd, because there is no _times_ argument anywhere to be found:

```
Quitting from lines 157-159 [unnamed-chunk-11] (double_dot_extension.Rmd)
   Error: processing vignette 'double_dot_extension.Rmd' failed with diagnostics:
   invalid 'times' argument
```

When I *knit* the rmd file, everything is fine. 

And I can run the code inside a function, so clearly the environment issue has been fixed here:

```
library(simstudy)

environment()
#> <environment: R_GlobalEnv>

test <- function() {
  
  print(environment())
  
  mult <- function(x) {
    2*x
  }
  
  defblk <- defData(varname = "blksize", 
                    formula = "mult(..sizes[1]) | .5 + mult(..sizes[2]) | .5", dist = "mixture")
  
  sizes <- c(2, 4)
  genData(1000, defblk)  
}

set.seed(123)
test()
#> <environment: 0x13b8f30e8>
#> Key: <id>
#>          id blksize
#>       <int>   <num>
#>    1:     1       4
#>    2:     2       8
#>    3:     3       4
#>    4:     4       8
#>    5:     5       8
#>   ---              
#>  996:   996       8
#>  997:   997       8
#>  998:   998       4
#>  999:   999       8
#> 1000:  1000       4
```

But, everything I have read indicates that this is possibly an environment issue with the devtools::check() environment. This makes sense since my changes have all been environment-related. But, I can't really understand the specific error message, nor have I been able to see where the environment is properly getting passed.

I'll keep looking, but this is the last remaining piece.

kgoldfeld:
I think I found the problem - there was some conflict with using the variable name ***n*** in `.evalWith` since we were using `list2evn`. There must be a a variable with that name in the devtools::check() environment. In any case,things checked OK on my laptop - hoping the tests here show the same thing. If they do, you should feel free to take a look and approve it seems OK. Thank goodness for all the tests.

R/generate_dist.R
```diff
@@ -455,7 +460,7 @@
 
 .getNBmean <- function(dtSim, formula, link, n = nrow(dtSim),
                        envir = parent.frame()) {
-  mean <- .evalWith(formula, .parseDotVars(formula, envir), dtSim, n)
+  mean <- .evalWith(formula, .parseDotVars(formula, envir), dtSim, n, envir)
   if (link == "log") {
     mean <- exp(mean)
   }
```
kgoldfeld, line 466: I am just testing my reviewing skills.

assignUser (approved):
Nice, looks good! 

> Thank goodness for all the tests.

Right? Tests give peace of mind :D In that regard it probably makes sense to add a test to make sure evaluation in non-global envs works for vars and functions?

kgoldfeld:
Yes - I'll add those tests. 

And also, I realize I am not totally comfortable with pulling in the entire environment - I would rather just pull in the functions. This way, if someone happens to have a variable that is the same as one that is used in the function, there is no conflict (as I experienced with the check environment). I am working on that, and will update  `.evalWith`.

kgoldfeld:
@assignUser A few things:

- I updated `.evalWith` to pass only function objects from the parent environment. This seems a little safer. Unless you think this is problematic, I will go ahead and merge.
- I received a message from the creator of package `gtsummary` that I need to "suggest" package `broom.helper` in order to use the package, which I do in at least one vignette. Indeed, a couple of days later, I received an e-mail from CRAN indicating that `simstudy` now has errors on [CRAN Package Check Results](https://cran.r-project.org/web/checks/check_results_simstudy.html). And those errors are related to calls to `gtsummary`. Which I guess highlights the downsides of using other packages that are evolving over time. Granted, this is only the second time in like 10 years (or however how old `simstudy` is) where I needed to make a change as a result of an external package change.
- Finally, the error was generated on [r-devel-linux-x86_64-fedora-clang](https://cran.r-project.org/web/checks/check_flavors.html#r-devel-linux-x86_64-fedora-clang). This led me to wonder how we determined which servers to check on the GitHub workflow. Shouldn't we be checking all the servers flavors that CRAN [checks](https://cran.r-project.org/web/checks/check_results_simstudy.html)?

R/internal_utility.R
```diff
@@ -105,14 +106,34 @@
...
+    function_objects <- all_objects[
+      sapply(
+        all_objects,
+        function(obj) {
+          is.function(get(obj, envir = envir))
+        })
+    ]
+    
+    functions_list <- mget(function_objects, envir = envir)
+    list2env(functions_list, envir = environment())
```
assignUser, line 130: I think you should add these to the same env used for the dot vars:
```suggestion
    list2env(functions_list, envir = e)
```

assignUser (approved):
Nice, looks good. One minor suggestion.

assignUser:
> Which I guess highlights the downsides of using other packages that are evolving over time.

Yep, it is always good to think about dependencies but you seem to have a good hand for that ;)

> This led me to wonder how we determined which servers to check on the GitHub workflow

I looked at the recommendations from the r-lib people: https://github.com/r-lib/actions/blob/v2-branch/examples/check-standard.yaml and modified those a bit.

Testing all flavors is not really worth it for a package without any complex build dependencies like simstudy as it can produce frequent errors that just vanish after a few days (for devel) and are a waste to investigate. In addition there is no real way to accuratly reproduce the cran environments as they don't provide any form of reproducible script and a lot of things are manually tweaked, so even if you try to test on these there is no gurantee that it will accuratley reflect the results you will get on CRAN (yes, this is from painful experience...).

And even if some real issue pops up on CRAN you are not constrained by any release process (compared to arrow for example) that slows you down in deploying a fix.

Looking at the current error  that happens on multiple versions and flavors I have no idea, haven't seen that one before. Looks like something broken with knitr? As it wasn't the case when initially submitted I am guessing this is on CRANs end.

R/internal_utility.R
kgoldfeld: I've made the change - that seems even safer. I will merge (and then submit to CRAN) once it passes our tests.

kgoldfeld:
> Looking at the current error that happens on multiple versions and flavors I have no idea, haven't seen that one before. Looks like something broken with knitr? As it wasn't the case when initially submitted I am guessing this is on CRANs end.

I am quite sure that the error is related to the changes in package `gtsummary` - which should go away since I have modified the *suggests* as they have recommended:

> I am preparing the submission of gtsummary v2.0 to CRAN, and I wanted to alert you to a change you may need to incorporate in simstudy.
> 
> The gtsummary package utilizes the broom.helpers package to prepare regression model results. In the updated version of the package, the broom.helpers package has moved from Imports to Suggests. As a result, you may also need to add broom.helpers to the Suggests field so the package is available when your vignettes are created.

PR #238, opened 2025-06-27
237 any possibility of adding narmfalse argument to the gendatadensity and adddatadensity functions

kgoldfeld:


kgoldfeld:
@assignUser Greetings - it has been a while. Hope you are well. I made a quick update to the `genDataDensity` and `addDataDensity` functions. I added an `na.rm` argument that allows users to generate missing data at the same rate that it appears in the distribution definition. I also added some related tests.

I've also made some updates to the workflow configuration - some versions were out of date. You don't need to review if you are too busy.

R/add_data.R
```diff
@@ -555,9 +557,12 @@ addDataDensity <- function(dtOld, dataDist, varname, uselimits = FALSE) {
   dataDist <- round(dataDist, 0)
   
   if (uselimits) {
-    density_est <- stats::density(dataDist, n = 10000, from = min(dataDist), to = max(dataDist))
+    density_est <- 
+      stats::density(dataDist, n = 10000, from = min(dataDist, na.rm = TRUE), 
+                     to = max(dataDist, na.rm = TRUE), na.rm = TRUE)
```
assignUser, line 562: This will set `na.rm` always to true here, given that it was always `FALSE` before I assume you'd want it to be based on the user input?

.github/workflows/touchstone-receive.yaml
assignUser, line 1: Good to move everything to recent versions! The only caveat is with the artifacts actions as some behaviors changed, but not how it's used here I'd say. :+1:

touchstone/config.json
assignUser, line 1: This was a frozen snapshot and now it's a moving target so it might cause issues through updates? But I guess it would be good to see those in CI anyway :)

.github/workflows/touchstone-receive.yaml
```diff
@@ -7,19 +7,17 @@ jobs:
...
-          content="${content//'%'/'%25'}"
-          content="${content//$'\n'/'%0A'}"
-          content="${content//$'\r'/'%0D'}"
-          # end of optional handling for multi line json
-          echo "::set-output name=config::$content"
+          {
+            echo "config<<EOF"
+            cat ./touchstone/config.json
+            echo "EOF"
+          } >> "$GITHUB_OUTPUT"
```
assignUser, line 19: Nice!

assignUser (approved):
One minor comment, looks great otherwise!

assignUser:
> it has been a while. Hope you are well. 
@kgoldfeld it has, thanks for the ping always happy to review! I've been well, hope you are too!

R/add_data.R
kgoldfeld: Yes - that is by design. I generate the data without missing data based on the non-missing data specification, and then if na.rm is set to FALSE, I randomly allocate missing values.

.github/workflows/touchstone-receive.yaml
kgoldfeld: Seems to have worked just fine. Without the updated versions, I was getting error messages.

.github/workflows/touchstone-receive.yaml
kgoldfeld: That's ChatGPT, not me. :)

PR #244, opened 2025-12-07
Fix mkl error

kgoldfeld:
Minor fix in `genCorMat` to address an issue where very small rounding differences could make correlation matrices appear non-symmetric, preventing some data generation. Matrices are now always properly symmetrized.


kgoldfeld:
@assignUser I had to do this tiny fix, because I was getting a CRAN error. I'd like to merge this, but giving you a chance to take look if you'd like. There was an error in one of the checks, but it had nothing to do with simstudy, so I am going to ignore it. Thanks.

